PR #9, opened 2026-04-11
fix: preserve redacted thinking blocks during reasoning cleanup

### LINKED ISSUE #8: Anthropic redacted_thinking blocks are stripped in stripClearedReasoning()
## Summary
When Magic Context transforms messages for Anthropic sessions, it can remove `redacted_thinking` parts from assistant messages. Anthropic requires those blocks to be replayed unchanged, so the request is rejected with errors like:

`messages.1.content.11: thinking or redacted_thinking blocks in the latest assistant message cannot be modified`

## Root cause
In `packages/plugin/src/hooks/magic-context/strip-content.ts`, `stripClearedReasoning()` includes `redacted_thinking` in the set of part types it may strip:

```ts
const CLEARED_REASONING_TYPES = new Set([thinking, reasoning, redacted_thinking]);
```

But the strip predicate only preserves parts that have a non-cleared `thinking` or `text` field:

```ts
const thinking = thinking in part ? (part.thinking as string | undefined) : undefined;
const text = text in part ? (part.text as string | undefined) : undefined;
return (
  (thinking !== undefined && thinking !== [cleared]) ||
  (text !== undefined && text !== [cleared])
);
```

That works for `thinking` / `reasoning`, but not for Anthropic `redacted_thinking` blocks, which do not use those fields. As a result, valid `redacted_thinking` parts fail the predicate and are removed.

## Why this looks like the bug
There is already a hint elsewhere in the code that `redacted_thinking` should *not* be treated as a normal mutable reasoning part:

`packages/plugin/src/hooks/magic-context/tag-content-primitives.ts`

```ts
export function isThinkingPart(part: unknown): part is ThinkingLikePart {
  if (part === null || typeof part !== object) return false;
  const candidate = part as Record<string, unknown>;
  return candidate.type === thinking || candidate.type === reasoning;
}
```

`isThinkingPart()` excludes `redacted_thinking`, but `stripClearedReasoning()` still targets it. That mismatch is what seems to break Anthropic replay.

## Minimal fix
Remove `redacted_thinking` from `CLEARED_REASONING_TYPES`:

```diff
- const CLEARED_REASONING_TYPES = new Set([thinking, reasoning, redacted_thinking]);
+ const CLEARED_REASONING_TYPES = new Set([thinking, reasoning]);
```

## Safer fix
Even with the one-line fix above, it would be good to make the preservation explicit:

```ts
if (partType === redacted_thinking) return true;
```

inside `stripClearedReasoning()` before the mutable reasoning checks.

## Extra hardening suggestion
There are other transform paths that clear adjacent assistant reasoning when tagged text or tool output is rewritten. For Anthropic sessions, it may be worth adding a provider-aware rule to never mutate the latest assistant message if it contains `thinking` or `redacted_thinking` parts.

## Environment where this surfaced
- `@cortexkit/opencode-magic-context@0.8.3`
- Anthropic models in OpenCode
- Error seen during transformed follow-up requests after assistant reasoning was present

If useful, I can also open a PR with the minimal patch.

tomolom:
## Summary
- stop `stripClearedReasoning()` from treating `redacted_thinking` as clearable reasoning content
- add a regression test proving assistant `redacted_thinking` blocks are preserved unchanged
- fixes #8

## Testing
- bun test packages/plugin/src/hooks/magic-context/strip-content.test.ts
- bun run typecheck

### DIFF
--- a/packages/plugin/src/hooks/magic-context/strip-content.test.ts
+++ b/packages/plugin/src/hooks/magic-context/strip-content.test.ts
@@ -193,6 +193,26 @@ describe("strip-content", () => {
                 });
             });
         });
+
+        describe("#given assistant messages with redacted thinking parts", () => {
+            describe("#when stripping cleared reasoning", () => {
+                it("#then preserves redacted thinking blocks unchanged", () => {
+                    const redactedPart = {
+                        type: "redacted_thinking",
+                        data: "opaque-provider-payload",
+                    };
+                    const textPart = { type: "text", text: "visible response" };
+                    const msg = message("m-1", "assistant", [redactedPart, textPart]);
+
+                    const stripped = stripClearedReasoning([msg]);
+
+                    expect(stripped).toBe(0);
+                    expect(msg.parts).toHaveLength(2);
+                    expect(msg.parts[0]).toBe(redactedPart);
+                    expect(msg.parts[1]).toBe(textPart);
+                });
+            });
+        });
     });
 
     describe("stripInlineThinking", () => {
--- a/packages/plugin/src/hooks/magic-context/strip-content.ts
+++ b/packages/plugin/src/hooks/magic-context/strip-content.ts
@@ -279,7 +279,7 @@ function findMaxTag(messageTagNumbers: Map<MessageLike, number>): number {
     return max;
 }
 
-const CLEARED_REASONING_TYPES = new Set(["thinking", "reasoning", "redacted_thinking"]);
+const CLEARED_REASONING_TYPES = new Set(["thinking", "reasoning"]);
 
 export function stripClearedReasoning(messages: MessageLike[]): number {
     let stripped = 0;


tomolom:
I considered making this provider-gated, but I think the unconditional fix is the safer shape here.

Reasoning:
- `stripClearedReasoning()` does not currently receive provider context.
- More importantly, `redacted_thinking` is already the provider-specific discriminator. If that part type is present, it is the opaque block that must be preserved rather than treated like clearable `thinking` / `reasoning` text.
- Leaving `redacted_thinking` in `CLEARED_REASONING_TYPES` means the cleanup path can delete a valid opaque block based only on missing `thinking` / `text` fields, which is wrong regardless of how provider detection is wired higher up.

If we want additional Anthropic-specific hardening later, that would make more sense one layer up in the transform pipeline (for example, skipping mutation of the latest assistant message when Anthropic reasoning blocks are present). But for this bug, removing `redacted_thinking` from the clearable-reasoning set seems like the correct local fix.

ualtinok:
Thanks! I never encountered redacted_thinking tag myself. Good catch!

PR #26, opened 2026-04-19
feat: auto-heal interceptor for strict provider prompt limits

Qiiks:
## Problem Statement
Certain LLM providers (specifically GitHub Copilot with `gpt-5.3-codex` and similar models) advertise a large context window (e.g., 400K) but enforce a much stricter **prompt/input limit** (e.g., 272K). 

While Magic Context v0.11.1 improved limit detection by preferring `limit.input`, standard reactive defenses (like the 95% block) are still vulnerable to **massive single-turn spikes** (e.g., pulling a 150K token Playwright DOM dump). In these cases:
1. The request hits the provider limit.
2. The provider returns a 400/422 error.
3. OpenCode panics and triggers its **built-in non-agentic compaction**, destroying the Magic Context session state.
4. The user is charged for a premium model request that failed.

## Proposed Solution: The Auto-Heal Interceptor
This PR introduces a `experimental.chat.messages.transform` interceptor that acts as a synchronous safety net right before the payload leaves the client.

### Key Features:
- **Synchronous Clipping:** If a request exceeds a user-defined `auto_heal_limit`, the plugin synchronously truncates the oldest tool outputs and raw text blocks until the payload fits safely under the ceiling.
- **Absolute Protection Zone:** It strictly protects the **newest 4 messages** (the current turn, newest tool results, and the active prompt) to ensure the model maintains its immediate instruction context.
- **Zero-Cost Background Healing:** Simultaneously triggers the Historian in the background using the configured free model to perform a permanent agentic compression for the next turn.
- **Zero Request Waste:** Prevents `ContextOverflowError` from ever hitting the network, saving user quota and preventing OpenCode from falling back to built-in compaction.

## Configuration
Added a new `auto_heal_limit` section to the schema (disabled by default):

```json
"auto_heal_limit": {
  "enabled": true,
  "default": 300000,
  "models": {
    "github-copilot/gpt-5.3-codex": 260000
  }
}
```

## Verification
- Added a new unit test suite in `transform.test.ts` (`autoHealLimit`).
- Verified that oldest tool outputs are clipped correctly.
- Verified that newest messages (the protection zone) remain 100% intact even during a truncation event.
- Verified that the Historian background flag is correctly signaled.

### DIFF
--- a/packages/plugin/src/config/schema/magic-context.ts
+++ b/packages/plugin/src/config/schema/magic-context.ts
@@ -119,6 +119,10 @@ export interface MagicContextConfig {
     iteration_nudge_threshold: number;
     history_budget_percentage: number;
     historian_timeout_ms: number;
+    auto_heal_limit?: {
+        enabled: boolean;
+        buffer_tokens: number;
+    };
     commit_cluster_trigger: {
         enabled: boolean;
         min_clusters: number;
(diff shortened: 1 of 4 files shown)


Qiiks:
Closing this PR in favour of a more targeted improvement. After reviewing the OpenCode source code (), I found that OpenCode already clamps all tool outputs (including MCP tools like Playwright) to 50KB / 2000 lines before they reach the context window. So the 150k token spike I was trying to defend against cannot actually happen via MCP tools — OpenCode truncates at the source. Additionally, v0.11.1 fixed the core issue by preferring `limit.input` (max_prompt_tokens = 272k) over `limit.context` (400k) for percentage calculations, so the 95% block now fires at the correct threshold (~258k tokens). A new PR will follow with a more focused improvement: making the tail-size trigger multiplier configurable for heavy tool-use workflows.

Issue #78, opened 2026-05-15
Embedding model fails to load in Electron - transformers.node.mjs ERR_MODULE_NOT_FOUND (v0.19.0+)


wwth8819:
## What happened
After upgrading from v0.17.0 to v0.19.0, the local embedding model (Xenova/all-MiniLM-L6-v2) fails to load on every 15-minute timer tick.

## Error log
[magic-context] embedding model failed to load: Cannot find module 'C:\Users\13096\.cache\opencode\packages\@cortexkit\opencode-magic-context@latest\node_modules\@huggingface\transformers\dist\transformers.node.mjs' imported from C:\Users\13096\.cache\opencode\packages\@cortexkit\opencode-magic-context@latest\node_modules\@cortexkit\opencode-magic-context\dist\index.js
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '...transformers.node.mjs'

## What I expected
Embedding model loads normally as it did in v0.17.0.

## Repro steps
1. Upgrade MC plugin from v0.17.0 to v0.19.0 via doctor --force + restart OpenCode Desktop
2. Check magic-context.log — embedding fails every 15 minutes on timer tick
3. The file transformers.node.mjs exists on disk at the specified path, but Node.js ESM resolver cannot find it

## Environment
- OpenCode Desktop v1.14.33 (Electron 34.3.0, Node ABI 145)
- Windows 11
- MC v0.19.0 (plugin cache)
- @huggingface/transformers v4.2.0
- Embedding config: local (Xenova/all-MiniLM-L6-v2 bundled)

## Notes
- Doctor check passes 14/14 (config-level check only)
- All other MC features work fine (transform, historian, dreamer, ctx_reduce)
- The file exists on disk — this appears to be an ESM module resolution issue specific to Electron environment

wwth8819:
[magic-context-issue-20260516-010152.md](https://github.com/user-attachments/files/27809813/magic-context-issue-20260516-010152.md)

ualtinok:
This should be fixed in v0.21.0, please let me know if the problem persists.

Number16BusShelter:
issue still seems to be present on version v0.21.6

```
[...]

PR #124, opened 2026-06-01
fix: inject WASM ONNX runtime for Bun to prevent NAPI fatal crash

### LINKED ISSUE #95: Bun/opencode crashes with NAPI fatal error when opencode-magic-context is enabled on macOS arm64
## Environment

- opencode: 1.15.10
- Bun: 1.3.14
- macOS: 26.4.1
- CPU: Apple Silicon / arm64
- Plugin: `@cortexkit/opencode-magic-context@latest`

## Symptom

opencode crashes after running for a while when `@cortexkit/opencode-magic-context@latest` is enabled.

The terminal crash output includes:

```text
panic: NAPI FATAL ERROR: Error::New napi_create_error
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
zsh: trace trap  opencode
```

## Evidence

The macOS crash report for the `opencode` process shows native modules loaded in the process:

```text
onnxruntime_binding.node
libonnxruntime.1.24.3.dylib
sharp-darwin-arm64.node
libvips-cpp.8.17.3.dylib
```

Those modules appear to come from the `@cortexkit/opencode-magic-context` dependency tree, specifically the ONNX / transformers and sharp / libvips native dependencies.

## Workaround

Removing `@cortexkit/opencode-magic-context@latest` from both opencode's global plugin list and TUI plugin list avoids loading those native modules. After removing it, opencode can load config and list models normally:

```text
opencode models newapi --print-logs
```

The startup logs no longer include `@cortexkit/opencode-magic-context@latest` and opencode successfully initializes the remaining plugins.

## Expected behavior

The plugin should not crash the host opencode / Bun process. If native dependencies are required, it would be helpful for the plugin to either avoid incompatible native paths under Bun or document a supported runtime / version matrix.

## Notes

This may ultimately be a Bun native-addon crash, but the practical trigger appears to be enabling `@cortexkit/opencode-magic-context@latest` in opencode on macOS arm64.

Suknna:
## Problem

opencode crashes with `NAPI FATAL ERROR: Error::New napi_create_error` when `@cortexkit/opencode-magic-context` is enabled on macOS arm64 (issue #95). The crash is triggered by `onnxruntime-node` (native N-API addon) hitting an incomplete N-API code path in Bun.

## Root cause

`@huggingface/transformers@4.x` statically imports `onnxruntime-node`. Under Bun, the native N-API binding calls `napi_create_error` which triggers Bun's fatal panic, crashing the entire host process. The crash is intermittent (happens "after running for a while") and depends on specific N-API error paths.

## Fix

Extend the existing WASM ONNX injection (`injectWasmOrtForElectron`, added for issue #78) to also activate under Bun. Both runtimes now get `onnxruntime-web` (WASM backend) instead of the native `onnxruntime-node`, while plain Node continues using the faster native backend.

**Detection:** `process.versions.bun` (Bun) || `process.versions.electron` (Electron) → inject WASM. Neither set (plain Node) → no-op.

## Verification

- ✅ `onnxruntime-web` loads and runs inference under Bun
- ✅ `Symbol.for("onnxruntime")` injection makes transformers.js use WASM backend
- ✅ Concurrent model loading works (5 parallel sessions)
- ✅ TypeScript typecheck passes
- ✅ Lint passes
- ✅ Build passes

## Trade-off

WASM backend is slower than native CPU, but it's the only reliable path under Bun. The native `onnxruntime-node` cannot safely load due to Bun's N-API limitations.

Closes #95

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/124"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782917537&installation_id=135454708&pr_number=124&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F124&signature=b2cf6454a080af9dfa0fa30f630da7b244ceb3a64cda390c609d5d0f2abfe230"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

---
## Summary by cubic
Switch Bun and Electron to `onnxruntime-web` (WASM) for embeddings to avoid `onnxruntime-node` N-API crashes. Plain Node keeps the native backend; closes #95.

- **Bug Fixes**
  - Injects WASM via `Symbol.for("onnxruntime")` when `process.versions.bun` or `process.versions.electron` is set; no-op on plain Node.
  - Prevents Bun panic (`NAPI FATAL ERROR: Error::New napi_create_error`) and Electron native load issues; falls back to native if injection fails.

<sup>Written for commit 2f58845d870674e0333521e3b6a3dc08e5eb2368. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/124?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/plugin/src/features/magic-context/memory/embedding-local.ts
+++ b/packages/plugin/src/features/magic-context/memory/embedding-local.ts
@@ -110,12 +110,11 @@ function startLockHeartbeat(lockPath: string): () => void {
  *
  * Why this exists:
  *   `@huggingface/transformers@4.x` does a top-level static `import "onnxruntime-node"`.
- *   On Electron Desktop (e.g. OpenCode Desktop's main process), that native
- *   `.node` binary fails to load for several environmental reasons — missing
- *   Visual C++ Redistributables on Windows, ASAR archive layout issues,
- *   onnxruntime's own dependency DLLs not being resolvable. The failure
- *   propagates up the import chain and Node reports it as `ERR_MODULE_NOT_FOUND`
- *   targeting `transformers.node.mjs`, even though that file exists.
+ *   On runtimes where the native `.node` binary fails to load — Electron
+ *   Desktop (missing Visual C++ on Windows, ASAR layout issues, ABI
+ *   mismatches) and **Bun** (incomplete N-API compatibility layer that
+ *   panics on `napi_create_error`, see issue #95) — the failure propagates
+ *   up the import chain and crashes the host process.
  *
  *   transformers.js exposes `Symbol.for("onnxruntime")` as an override hook
  *   (added in 3.4.x via PR #1231). If that global symbol is set before the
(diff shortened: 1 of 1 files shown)


alfonso-magic-context:
Thanks for digging into this and tracing it to `napi_create_error` — the diagnosis is solid.

We're going to hold off on this one, though. Forcing the WASM ONNX backend on Bun would regress *every* terminal OpenCode user (terminal OpenCode runs under Bun) from native-CPU embeddings to the slower WASM path, and the NAPI fatal is fundamentally a Bun N-API compatibility gap. We'd rather not take a permanent perf regression on the common runtime for a bug that belongs upstream — we're waiting on the next Bun release, which should fix the incomplete N-API path that triggers this.

Closing for now. If a Bun release lands and #95 still reproduces, we'll revisit. Really appreciate the careful write-up.

PR #141, opened 2026-06-12
fix: historian livelock starves pending drops in active sessions

tracycam:
## Problem

In an active session, the historian can livelock and pending drop operations starve forever. Observed in production (OpenCode harness): a session accumulated 179 pending drops while active context grew past 280K tokens with zero reduction.

The cycle, repeating every turn:

1. commit-cluster trigger fires (clusters never get compartmentalized, so it refires every turn)
2. `compartmentInProgress` flag set, historian agent starts
3. `historian no-op: stale protected-tail snapshot (last ordinal N id changed)` — the boundary snapshot is captured at trigger time, and the ONLY failing validation is the "last message in protected tail" tuple, which changes every turn in an active session
4. `transform: deferring pending ops — compartment agent in progress` — the no-op returns synchronously, but the `activeRuns` registration is only cleared via a microtask `promise.finally`, so the same pass still believes a run is active and defers pending ops

## Fix (two independent defects)

1. **Re-derive the boundary snapshot at run time on staleness.** On a `stale_snapshot` validation result, the runner re-resolves the boundary once from current session state and adopts it if it still has a runnable head. `validateBoundarySnapshot` itself is untouched — the refreshed snapshot recomputes `protectedTailStart`/`eligibleEnd` from live messages, so the compacted head can never include a message that belongs to the live protected tail.
2. **Synchronous no-op clears the active-run registration.** `startCompartmentAgent` now registers the run at the real-run commit point (after the last synchronous no-op return, before the first await) via `onHistorianRunStarted`, and drops the registration synchronously when the runner no-ops — so pending ops can materialize in the same pass.

## Tests

Two regression tests in `compartment-runner.test.ts`:
- refreshes a stale-tail boundary snapshot at run time instead of no-op'ing forever (asserts publish + queued drops + `compartmentInProgress=false`)
- clears the active-run registration synchronously when the runner no-ops

`bun run typecheck`, `bun run lint`, full plugin suite pass (the two pre-existing flaky timing tests aside).

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/141"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783885617&installation_id=135454708&pr_number=141&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F141&signature=98e6f399c00dc661a9104c3c2c65db1234103345b33f8f20757f20133a6c22d0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Fixes a historian livelock that starved pending drop ops in active sessions. The runner now refreshes stale boundaries and clears no-op runs immediately, so compaction proceeds and drops drain.

- **Bug Fixes**
  - Re-resolves the protected-tail boundary on `stale_snapshot` at run time and adopts it only if the head is runnable, preserving protected-tail safety.
  - Clears active-run registration synchronously on no-op via `onHistorianRunStarted`, so `activeRuns` doesn’t block pending drops in the same pass.

<sup>Written for commit 055da42fb6822c0ad0f953f599d1500c3e5a7814. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/141?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

This PR fixes a production livelock where the historian agent repeatedly no-ops due to a stale boundary snapshot while pending drop operations starve indefinitely. Two independent defects are addressed: (1) on a `stale_snapshot` validation result the runner now re-resolves the boundary once from live DB state and proceeds if a runnable window still exists, and (2) the `activeRuns` registration is cleared synchronously when the runner determines it will no-op, so the same transform pass can drain pending drops instead of deferring them.

- **Stale-snapshot refresh** (`compartment-runner-incremental.ts`): when `validateBoundarySnapshot` returns `reason: \"stale_snapshot\"`, the runner calls `resolveOpenCodeProtectedTailBoundary` against the live DB and adopts the refreshed snapshot if `hasRunnableCompartmentWindow` is satisfied. The runner's own `eligibleEndOrdinal = Math.min(snapshot.eligibleEndOrdinal, protectedTailStart)` clamp maintains the protected-tail guarantee regardless of snapshot staleness.
- **Synchronous `activeRuns` cleanup** (`compartment-runner.ts`): `startCompartmentAgent` introduces a `realRunStarted` boolean set by the new `onHistorianRunStarted` callback, called just before the first real `await`. After `runCompartmentAgent` is launched, if `realRunStarted` is still false the registration is deleted synchronously before the microtask-scheduled `.finally` can clear it — unblocking pending-op drain in the same pass.
</details>

Safe to merge — the two independent fixes address a well-documented production livelock with targeted, narrow changes backed by regression tests.

Both changes are tightly scoped and the core safety guarantee (protected tail never compacted) is enforced by the runner's own Math.min(eligibleEndOrdinal, protectedTailStart) clamp independent of the refreshed snapshot. The realRunStarted mechanism is correct for both synchronous and async no-op paths. The two findings are documentation and clarity gaps, not behavioral defects.

The stale-snapshot refresh block in compartment-runner-incremental.ts (lines 244–266) and the matching onHistorianRunStarted comment in both compartment-runner-incremental.ts and compartment-runner.ts would benefit from a one-line clarification about async early-return coverage and stale usage forwarding.

| Filename | Overview |
|----------|----------|
| packages/plugin/src/hooks/magic-context/compartment-runner.ts | Introduces realRunStarted flag + onHistorianRunStarted callback to synchronously clear activeRuns when the runner no-ops; lease and interval cleanup still run via .finally() microtask. Logic is sound for both synchronous and async no-op paths. |
| packages/plugin/src/hooks/magic-context/compartment-runner-incremental.ts | Adds stale-snapshot refresh: on stale_snapshot validation failure, re-resolves boundary from live DB using trigger-time usage/threshold params and adopts it when hasRunnableCompartmentWindow passes. Protected-tail safety is preserved by the runner's own Math.min(eligibleEndOrdinal, protectedTailStart) clamp downstream. |
| packages/plugin/src/hooks/magic-context/compartment-runner-types.ts | Adds optional onHistorianRunStarted callback to CompartmentRunnerDeps with clear JSDoc; no breaking changes. |
| packages/plugin/src/hooks/magic-context/compartment-runner.test.ts | Adds two targeted regression tests: stale-tail refresh (verifies publish + pending ops + compartmentInProgress=false) and sync-noop activeRuns cleanup (synchronous assertion before any await). Both cover the exact production failure modes described in the PR. |

</details>

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant T as Transform Pass
    participant SCA as startCompartmentAgent
    participant RCA as runCompartmentAgent
    participant DB as DB / State

    T->>SCA: startCompartmentAgent(deps)
    SCA->>DB: acquireCompartmentLease()
    SCA->>RCA: runCompartmentAgent(runnerDeps) [fire-and-forget]

    alt No-op path (stale snapshot)
        RCA->>DB: validateBoundarySnapshot → stale_snapshot
        Note over RCA: NEW: re-resolve boundary
        RCA->>DB: resolveOpenCodeProtectedTailBoundary (live)
        alt hasRunnableCompartmentWindow
            RCA->>RCA: "boundarySnapshot = refreshed, validation.ok = true"
            Note over RCA: continues to real run
            RCA-->>SCA: onHistorianRunStarted() called
            Note over SCA: realRunStarted = true
            RCA->>DB: publish compartments
        else no runnable window
            RCA-->>RCA: rollback, return (no await reached)
            Note over SCA: realRunStarted = false
        end
    else No-op path (nothing to compact)
        RCA-->>RCA: rollback, return synchronously
        Note over SCA: realRunStarted = false
    end

    SCA->>SCA: activeRuns.set(sessionId, promise)
    Note over SCA: NEW: if (!realRunStarted) activeRuns.delete() synchronously
    SCA-->>T: returns

    T->>T: getActiveCompartmentRun() → undefined
    T->>DB: drain pending drops (no longer blocked)

    Note over RCA: microtask later: .finally() clears interval + lease
```
</details>

<sub>Reviews (1): Last reviewed commit: ["plugin: refresh protected-tail snapshot ..."](https://github.com/cortexkit/magic-context/commit/055da42fb6822c0ad0f953f599d1500c3e5a7814) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37303454)</sub>

> Greptile also left **2 inline comments** on this PR.

### DIFF
--- a/packages/plugin/src/hooks/magic-context/compartment-runner-incremental.ts
+++ b/packages/plugin/src/hooks/magic-context/compartment-runner-incremental.ts
@@ -56,7 +56,9 @@ import { clearInjectionCache, renderMemoryBlock } from "./inject-compartments";
 import { onNoteTrigger } from "./note-nudger";
 import {
     createDefaultBoundarySnapshotForTests,
+    hasRunnableCompartmentWindow,
     recordHighPressureNoEligibleHead,
+    resolveOpenCodeProtectedTailBoundary,
     selectPerRunCap,
     validateBoundarySnapshot,
 } from "./protected-tail-boundary";
@@ -203,7 +205,7 @@ export async function runCompartmentAgent(deps: CompartmentRunnerDeps): Promise<
                 ? priorCompartments[priorCompartments.length - 1].endMessage + 1
                 : 1;
 
-        const boundarySnapshot =
+        let boundarySnapshot =
             deps.boundarySnapshot ??
             (process.env.NODE_ENV === "test"
                 ? createDefaultBoundarySnapshotForTests(sessionId)
@@ -217,7 +219,7 @@ export async function runCompartmentAgent(deps: CompartmentRunnerDeps): Promise<
             rollbackDrainReservation();
             return;
         }
-        const validation =
+        let validation =
             boundarySnapshot.rawRangeFingerprint.length > 0
                 ? validateBoundarySnapshot({
                       db,
@@ -226,6 +228,42 @@ export async function runCompartmentAgent(deps: CompartmentRunnerDeps): Promise<
                           deps.currentContextLimit ?? boundarySnapshot.contextLimit,
                   })
                 : { ok: true };
+        // In an active session the protected tail's newest message changes every
+        // turn (a fresh user/assistant message lands), so a snapshot captured at
+        // trigger time goes stale on the "last ordinal id" check by the time the
+        // historian actually runs — even though the ELIGIBLE HEAD (offset →
+        // eligibleEnd) it would compact is untouched. Left as-is the runner no-ops
+        // forever while the trigger refires each turn, and queued drop ops starve
+        // (observed in production: 27 consecutive stale-snapshot no-ops, 179
+        // pending drops, zero reduction). Re-resolve the boundary ONCE from the
+        // CURRENT session state and adopt the fresh snapshot when it still exposes
+        // a runnable head. This does NOT weaken the protected-tail guarantee: the
+        // refreshed snapshot recomputes protectedTailStart/eligibleEnd from the
+        // live messages, so the head can never include a message that now belongs
+        // to the current protected tail.
+        if (!validation.ok && validation.reason === "stale_snapshot") {
+            const refreshed = resolveOpenCodeProtectedTailBoundary({
+                db,
+                sessionId,
+                mode: "incremental-runner",
+                contextLimit: deps.currentContextLimit ?? boundarySnapshot.contextLimit,
+                executeThresholdPercentage: boundarySnapshot.executeThresholdPercentage,
+                usage: {
+                    percentage: boundarySnapshot.usagePercentage,
+                    inputTokens: boundarySnapshot.usageInputTokens,
+                },
+                usageSource: boundarySnapshot.usageSource,
+                emergencyTailScale: boundarySnapshot.emergencyTailScale,
+            });
+            if (hasRunnableCompartmentWindow(refreshed)) {
+                sessionLog(
+                    sessionId,
+                    `historian: refreshed stale protected-tail snapshot at run time (was: ${validation.detail ?? "stale"}) — eligible head ${refreshed.offset}-${refreshed.eligibleEndOrdinal - 1}`,
+                );
+                boundarySnapshot = refreshed;
+                validation = { ok: true };
+            }
+        }
         if (!validation.ok) {
             sessionLog(
                 sessionId,
@@ -334,6 +372,14 @@ export async function runCompartmentAgent(deps: CompartmentRunnerDeps): Promise<
             rollbackDrainReservation();
             return;
         }
+
+        // Past every synchronous no-op early-return and immediately before the
+        // first `await` (client.session.get below): we are now committed to a
+        // real historian pass. Signal the caller so startCompartmentAgent keeps
+        // the active-run registration; a synchronous no-op above never reaches
+        // here, so its lingering registration is cleared instead of blocking the
+        // same transform pass's pending-op drain.
+        deps.onHistorianRunStarted?.();
 
         // v2 bounded reference model (replaces the unbounded existing_state dump):
         //   - 4 rotating cross-project seeds + last-6 recency compartments (no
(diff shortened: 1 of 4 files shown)


alfonso-magic-context:
Merged into master (cherry-picked as `cdd17f8`, your authorship preserved). Thanks — this is a clean diagnosis and fix of a real production starvation path.

We independently traced the livelock against current source and confirmed it: `runCompartmentAgent` returns synchronously on the stale-snapshot no-op, while `activeRuns` only clears in the microtask `.finally`, so the same transform pass still sees a run "active" and `shouldApplyPendingOps` defers the drops — and the commit-cluster trigger refires every turn, so it never recovers. Both halves of your fix are right: re-resolving the boundary at run time, and clearing the registration synchronously on a no-op.

On the bot review threads: the async early-return paths (`existingValidationError` / `chunkCoverageError` await a notification before `onHistorianRunStarted`) are not a duplicate-run race — they don't publish compartment state, and the compartment lease stays held until `finally`, so a second start can't acquire it. The trigger-time `usagePercentage` forwarded into the run-time re-resolve only makes the refreshed protected tail *more conservative*; the runner still clamps `eligibleEndOrdinal` to `protectedTailStart`, so nothing past the live tail is ever compacted. No changes needed.

Full plugin suite green locally including your two regression tests.

tracycam:
Thanks for the clean merge, and for tracing the livelock independently against current source before taking it — that's more diligence than PRs usually get. Glad the authorship survived the cherry-pick. Your read on the bot-flagged paths matches mine: the async early-returns never release the compartment lease before `finally`, so the duplicate-start race can't materialize, and the run-time re-resolve only makes the protected tail more conservative. Nothing further from my side.

PR #142, opened 2026-06-12
fix: restore message snapshot when tagging fails mid-pass

tracycam:
## Problem

`tagMessages()` mutates message parts in place (e.g. `§N§` text prefixing) after each DB write. If it throws mid-pass (DB error, etc.), the emitted `messages` array keeps the partial mutations — the LLM silently receives a prompt with partial tag prefixes that don't correspond to persisted state.

## Fix

Snapshot `{ info, parts }` (structuredClone) immediately before `tagMessages()`; on the tagging-failure catch path, restore via `messages.splice(0, messages.length, ...snapshot)`. The existing `taggingSucceeded` gate and the documented no-outer-transaction design of `tagMessages()` are untouched — this only rolls back in-memory mutations, not durable DB semantics.

## Tests

- mid-pass failure: tagger throws on the 2nd message after the 1st was already prefixed → asserts no part contains a `§N§` prefix and both messages match their pre-tag text
- happy path: successful pass still persists `source_contents` and replays dropped statuses (snapshot/restore is a no-op on success)

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/142"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783885629&installation_id=135454708&pr_number=142&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F142&signature=d9e278c9d4fdc39ee485bf3af2f89a5561636047869c15d5101e15b2ec25baf3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

This PR guards against partial in-memory corruption when `tagMessages()` throws mid-walk by taking a `structuredClone` snapshot of `{info, parts}` immediately before the call and splicing the snapshot back on the catch path. The existing `taggingSucceeded` gate and no-outer-transaction design of `tagMessages` are left intact.

- `snapshotMessagesForTagging` clones every message's `info` and `parts` before tagging begins; on failure the `messages` array contents are restored so no stray `§N§` prefixes reach the LLM.
- Two new integration tests cover the rollback path (tagger forced to throw on the 2nd `assignTag`, asserts all text reverts to pre-tag content) and the happy path (source_contents and dropped-status replay still work correctly after the snapshot code was added).
</details>

Safe to merge; the fix correctly restores the message array on the failure path and is well-tested, with two minor robustness items worth a follow-up.

The snapshot/restore logic is sound and the taggingSucceeded gate prevents downstream stages from running on stale data. Two items are worth noting: structuredClone is called outside the try-catch so a DataCloneError from a non-cloneable part would bubble up as an unhandled transform failure, and the ...preTagSnapshot spread into splice has a theoretical argument-count ceiling. Neither is likely to trigger in practice given JSON-like parts and bounded message windows.

packages/plugin/src/hooks/magic-context/transform.ts — specifically the placement of snapshotMessagesForTagging relative to the try-catch and the splice restore pattern.

| Filename | Overview |
|----------|----------|
| packages/plugin/src/hooks/magic-context/transform.ts | Adds snapshotMessagesForTagging helper and a messages.splice restore on the tagging catch path; snapshot is taken outside the try-catch (structuredClone failure would be unhandled) and the spread-into-splice pattern has a theoretical argument-count ceiling. |
| packages/plugin/src/hooks/magic-context/transform.test.ts | Adds two new integration tests: mid-pass rollback (tagger throws on 2nd assignTag, asserts no §N§ prefixes survive) and happy-path regression (source_contents persisted, dropped status replayed correctly after snapshot code was added). |

</details>

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant T as transform()
    participant S as snapshotMessagesForTagging()
    participant TM as tagMessages()
    participant DB as ContextDatabase

    T->>S: "structuredClone({info, parts}) for each message"
    S-->>T: preTagSnapshot[]

    T->>TM: tagMessages(sessionId, messages, tagger, db)
    loop per message part
        TM->>DB: tagger.assignTag() SAVEPOINT
        DB-->>TM: tag number
        TM->>TM: prependTag mutates part.text in place
    end

    alt tagging succeeds
        TM-->>T: TagMessagesResult
        T->>T: "taggingSucceeded = true"
        Note over T: downstream stages run normally
    else tagMessages throws mid-walk
        TM--xT: Error
        Note over T: messages[0] already has §1§ prefix
        T->>T: restore via splice
        Note over T: in-memory array restored to pre-tag state
        T->>T: tagger.cleanup(sessionId)
        Note over T: taggingSucceeded stays false
    end
```
</details>

<sub>Reviews (1): Last reviewed commit: ["plugin: restore message snapshot when ta..."](https://github.com/cortexkit/magic-context/commit/ef0f13b554e0cd6b6ef2ff854f13f57c5ed77c8c) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37303487)</sub>

> Greptile also left **2 inline comments** on this PR.

---
## Summary by cubic
Restores the pre-tag message snapshot when `tagMessages()` fails mid-pass, preventing partial `§N§` prefixes from leaking into prompts. Keeps LLM input consistent with persisted state.

- **Bug Fixes**
  - Take a deep snapshot of `{ info, parts }` before tagging and splice it back on failure.
  - Keep `taggingSucceeded` gating and DB semantics unchanged; only in-memory mutations are rolled back.
  - Add integration tests for mid-pass rollback and the happy path (persisting `source_contents` and replaying dropped status).

<sup>Written for commit ef0f13b554e0cd6b6ef2ff854f13f57c5ed77c8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/142?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/plugin/src/hooks/magic-context/transform.test.ts
+++ b/packages/plugin/src/hooks/magic-context/transform.test.ts
@@ -23,6 +23,7 @@ import {
     getLastNudgeUndropped,
     getOrCreateSessionMeta,
     getPendingOps,
+    getSourceContents,
     getTagById,
     getTagsBySession,
     incrementHistorianFailure,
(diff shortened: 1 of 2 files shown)


alfonso-magic-context:
Thanks for this — the failure mode it guards against is real.

Our one concern is the hot path: `structuredClone` of the entire messages array on **every** transform pass is real per-turn overhead (the transform runs on every request, and the array can be large in long sessions) to protect against a rare mid-pass tagging failure. We're sensitive to hot-path allocation cost here.

Could you rework it so the happy path stays allocation-free — e.g. clone only on the failure path (catch → restore from a snapshot taken lazily), or capture a cheap restore token rather than a full deep clone up front? A version that adds no per-pass cost when tagging succeeds we'd be glad to merge. Happy to discuss the approach if helpful.

alfonso-magic-context:
Thanks for this — the failure mode you identified is real and important: if `tagMessages()` throws mid-walk, the partially `§N§`-prefixed array can reach the model. We want that closed.

We're going a different route than the per-pass `structuredClone` snapshot, though. The transform runs on the hot path on every LLM call, and cloning the full message array each pass to guard a rare failure adds overhead we'd rather not pay on the 99.99% success path. Instead the direction we've settled on is:

1. Build the managed array in private/local state and assign `output.messages` only on success — so a mid-transform throw never leaves a half-mutated array in the first place (no snapshot needed).
2. On genuine unrecoverable failure, fall back to replaying the last-known-good cached m[0]/m[1] (boundary-validated) + a bounded tail-trim, rather than passing raw/partial history through.

That work folds into a larger transform/cache refactor we have queued, so I'm closing this PR rather than merging the clone approach — but you surfaced the exact bug it needs to handle, and that shaped the design. Thank you.

tracycam:
Ack — and I think your route is the better structure. The per-pass structuredClone was paying hot-path cost on every transform to defend against a mutation that shouldn't be observable mid-flight at all; build-then-assign gives the same guarantee (a mid-pass throw can never leave a half-mutated array) by construction instead. I can see the direction already in master: the tagging failure path records `tagging-persistence-failure`, drops in-memory tagger state so the next pass self-heals, and the downstream mutation stages are gated on tagging success so a throw can't fan out partial work. Would appreciate a tag when the full transform/cache refactor (build-then-assign + last-known-good m0/m1 replay) lands — curious how the boundary validation works on the replay path.

PR #169, opened 2026-06-18
fix(tui): null-guard RecompProgressSection against reactive race (#168)

### LINKED ISSUE #168: TUI fatal crash: null is not an object (evaluating 'props.progress.phase') in RecompProgressSection
## What happened?

The OpenCode TUI crashed fatally while the magic-context sidebar was rendering the recomp / session-upgrade progress section. The error toast surfaced this stack:

```
Error: null is not an object (evaluating 'props.progress.phase')
    at phase (.../src/tui/slots/sidebar-content.tsx:376:29)
    at <anonymous> (.../src/tui/slots/sidebar-content.tsx:491:38)
    ...
```

After the crash, the TUI process is gone and the session must be restarted.

## Root cause

Two render sites in `packages/plugin/src/tui/slots/sidebar-content.tsx` (the collapsed-view and the expanded-view render of `RecompProgressSection`) use a fragile Solid JSX pattern:

```tsx
{s()?.recompProgress && (
    <RecompProgressSection theme={props.theme} progress={s()!.recompProgress!} />
)}
```

The guard and the prop expression read the same signal twice:

1. **Guard** uses optional chaining (`s()?.recompProgress`) — admits `null | undefined`.
2. **Prop** uses non-null assertion (`s()!.recompProgress!`) — assumes non-null.
3. In Solid, the prop expression `progress={s()!.recompProgress!}` is re-evaluated on every reactive access of `props.progress` inside the child (this reactivity is intentional — the original code comment at L356-L363 explains that the child MUST observe phase transitions live without remounting).
4. The `recompTick` resilient poll loop (L540-L606) explicitly handles transient cache misses by carrying the last good progress forward, but when both the current snapshot AND `prevProgress` are falsy, `setSnapshot({ ...data, recompProgress: null })` is published.
5. On the next reactive read inside `RecompProgressSection`, `props.progress` evaluates `s()!.recompProgress!` → `null` → `props.progress.phase` throws `TypeError: null is not an object`.

The `NonNullable<SidebarSnapshot["recompProgress"]>` annotation on the component's `progress` prop is a compile-time promise that nothing enforces at runtime — and the repo's default `bun run typecheck` (tsconfig `include: ["src/**/*.ts"]`) and `bun run lint` (biome `includes: ["src/**/*.ts"]`) don't cover `.tsx` files, so this null-safety hole slipped through CI.

## Expected behavior

The sidebar should never crash on a transient `recompProgress: null` snapshot. It should silently skip rendering the progress section until the next poll returns a non-null value.

## Fix

Replace both render sites with the canonical Solid `<Show>` callback form, which guarantees the child callback only runs while `when` is truthy and the accessor returns the narrowed non-null value:

```tsx
<Show when={s()?.recompProgress}>
    {(progress) => (
        <RecompProgressSection theme={props.theme} progress={progress()} />
    )}
</Show>
```

This also removes two `noNonNullAssertion` lint warnings.

PR: <filling in>

## Repro

Hard to repro deterministically — requires a recomp/upgrade pass to coincide with a snapshot cache miss that publishes `{ recompProgress: null }` while `prevProgress` is also falsy. Easiest path: start a `/ctx-recomp`, then rapidly issue concurrent snapshot fetches until one returns a fresh (no recompProgress) entry.

## Diagnostics

- Plugin version: 0.25.0
- OpenCode version: 1.17.8
- Platform: macOS arm64
- Client: OpenCode TUI (CLI)

## Suggested follow-up (out of scope for this fix)

`packages/plugin/biome.json` and `packages/plugin/tsconfig.json` both scope their includes to `src/**/*.ts` and exclude `.tsx`. Type and lint errors in `.tsx` files are not caught by `bun run typecheck` / `bun run lint` in CI. Adding `src/**/*.tsx` to both include lists would prevent a similar class of bug from shipping.

wjiuxing:
## Summary

Fixes #168 — fatal TUI crash `null is not an object (evaluating 'props.progress.phase')` in `RecompProgressSection`.

Replaces two fragile Solid JSX render sites with the canonical `<Show>` callback form, eliminating a reactive null-deref race that crashed the TUI when the sidebar snapshot transiently published `recompProgress: null` during a recomp pass.

## Root cause

Both render sites of `RecompProgressSection` used this pattern:

```tsx
{s()?.recompProgress && (
    <RecompProgressSection theme={props.theme} progress={s()!.recompProgress!} />
)}
```

The guard reads `s()` with optional chaining (admits null), but the component prop reads `s()!` with non-null assertion (assumes non-null). In Solid, `progress={s()!.recompProgress!}` is re-evaluated on every reactive access of `props.progress` inside the child — this reactivity is intentional (the L356–L363 comment explains the child must observe phase transitions live without remounting). When `setSnapshot({ ...data, recompProgress: null })` is published during a transient cache miss (the resilient poll loop at L540–L606 carries the last good progress forward, but when both current snapshot AND `prevProgress` are falsy, it publishes a null), the next reactive read of `props.progress.phase` dereferences null and crashes.

The `NonNullable<SidebarSnapshot["recompProgress"]>` annotation on the component's `progress` prop is a compile-time promise that nothing enforces at runtime.

## The fix

```tsx
<Show when={s()?.recompProgress}>
    {(progress) => (
        <RecompProgressSection theme={props.theme} progress={progress()} />
    )}
</Show>
```

`<Show>`'s callback form guarantees:
1. The child callback only runs while `when` is truthy.
2. The `progress()` accessor returns the narrowed non-null value reactively.
3. When `when` goes falsy, the callback unmounts — no further reactive accesses to `props.progress` are possible.

Also removes two `noNonNullAssertion` lint warnings.

## Files changed

- `packages/plugin/src/tui/slots/sidebar-content.tsx`
  - Added `Show` to the `solid-js` import.
  - Replaced the collapsed-view render site (L758–L762).
  - Replaced the expanded-view render site (L794–L798).

## Scope

- **In scope:** Eliminate the null-deref crash in the two known render sites of `RecompProgressSection`.
- **Out of scope (noted in #168):** `packages/plugin/biome.json` and `packages/plugin/tsconfig.json` both scope their includes to `src/**/*.ts` and exclude `.tsx`, so type and lint errors in `.tsx` files are not caught by `bun run typecheck` / `bun run lint` in CI. Adding `src/**/*.tsx` to both include lists would prevent this class of bug from shipping. Left for a separate PR per the repo's scope preference.

## Test plan

- [x] `bun run typecheck` — exit 0 (default scope)
- [x] `bun run lint` — exit 0, 494 files checked (default scope)
- [x] `bun run build` — exit 0, 374 modules bundled (bun's transpiler covers `.tsx`)
- [x] `bun run test` — 2103 pass / 0 fail (baseline: 2103 pass / 0 fail)
- [ ] Manual: trigger a `/ctx-recomp` and confirm the progress section still renders live updates without crashing across snapshot cache misses.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/169"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784355925&installation_id=135454708&pr_number=169&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F169&signature=faf34474b1ae2f3f990e0a4e6e4f3658234bbd3fab01defe99d991bd22019d80"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

---
## Summary by cubic
Prevents a TUI crash by switching `RecompProgressSection` to Solid’s <Show> callback so it only renders when `recompProgress` exists. This removes the reactive null-deref that occurred during transient cache misses.

- **Bug Fixes**
  - Replaced both sidebar render sites with <Show when={s()?.recompProgress}>{(progress) => <RecompProgressSection ... progress={progress()} />}</Show> to avoid null derefs and auto-unmount when falsy.
  - Imported `Show` from `solid-js` and removed non-null assertions on the `progress` prop.

<sup>Written for commit a87fa703fe9eb5473e7907490dbf2d371662d09e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/169?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/plugin/src/tui/slots/sidebar-content.tsx
+++ b/packages/plugin/src/tui/slots/sidebar-content.tsx
@@ -1,5 +1,5 @@
 /** @jsxImportSource @opentui/solid */
-import { createEffect, createMemo, createSignal, on, onCleanup } from "solid-js"
+import { Show, createEffect, createMemo, createSignal, on, onCleanup } from "solid-js"
 import type { TuiSlotPlugin, TuiPluginApi, TuiThemeCurrent } from "@opencode-ai/plugin/tui"
 import packageJson from "../../../package.json"
 import { loadSidebarSnapshot, type SidebarSnapshot } from "../data/context-db"
@@ -755,9 +755,11 @@ const SidebarContent = (props: {
                             C:{s()?.compartmentCount ?? 0} Q:{s()?.pendingOpsCount ?? 0} N:{s()?.sessionNoteCount ?? 0}
                         </text>
                     </box>
-                    {s()?.recompProgress && (
-                        <RecompProgressSection theme={props.theme} progress={s()!.recompProgress!} />
-                    )}
+                    <Show when={s()?.recompProgress}>
+                        {(progress) => (
+                            <RecompProgressSection theme={props.theme} progress={progress()} />
+                        )}
+                    </Show>
                 </box>
             )}
 
@@ -789,9 +791,11 @@ const SidebarContent = (props: {
             />
 
             {/* Recomp / session-upgrade live progress */}
-            {s()?.recompProgress && (
-                <RecompProgressSection theme={props.theme} progress={s()!.recompProgress!} />
-            )}
+            <Show when={s()?.recompProgress}>
+                {(progress) => (
+                    <RecompProgressSection theme={props.theme} progress={progress()} />
+                )}
+            </Show>
                 </>
             )}
 


alfonso-magic-context:
Thank you @wjiuxing! Merged into master as `58475a7e` (cherry-picked, so this PR won't auto-close) and shipped in **v0.26.0**. The `<Show>` callback form is exactly right — the prop re-evaluated reactively and dereferenced the transient null. Credited you as co-author on the commit.

PR #292, opened 2026-08-09
fix(prompt-surface): let the generator own the A1 golden's hash baseline

### LINKED ISSUE #290: export-agent-surface.ts omits section 3, so regenerating the A1 golden breaks the guidance-epoch test
`packages/plugin/src/shared/prompt-surface-a1-golden.md` says it is generated by `export-agent-surface.ts`, but that script emits only sections 1 and 2. Section 3 (the system-prompt hash baseline table) is hand-maintained. So running the script the way its own usage line documents silently deletes a section the test suite depends on.

Reproduced on clean `upstream/master` @ `77c51da9`, fresh worktree + `bun install`:

```
$ grep -c '^## 3\. System-prompt hash baseline' packages/plugin/src/shared/prompt-surface-a1-golden.md
1
$ bun packages/plugin/scripts/export-agent-surface.ts packages/plugin/src/shared/prompt-surface-a1-golden.md
written: .../prompt-surface-a1-golden.md          # exit 0, no warning
$ grep -c '^## 3\. System-prompt hash baseline' packages/plugin/src/shared/prompt-surface-a1-golden.md
0
$ bun test packages/plugin/src/hooks/magic-context/system-prompt-hash.test.ts
error: Malformed A1 primary guidance golden
(fail) OpenCode prompt-surface guidance epochs > matches A1 guidance for no config and explicit full
```

`readA1PrimaryGuidance` (`system-prompt-hash.test.ts:854`) matches `| PRIMARY full | \d+ | \`([0-9a-f]{32})\` |` and throws when it is absent.

Two more consumers slice on that heading as an end delimiter — `tool-registry.test.ts:213` and `pi-plugin/src/tools/index.test.ts:252` both do `document.slice(document.indexOf("## 2. Tool surface"), document.indexOf("## 3. System-prompt hash baseline"))`. With the heading gone that second argument is `-1`, so the slice silently returns `""` and the golden parses as zero tools. Those two happen to still pass today (empty-vs-empty), which makes it worse: the only signal is the one thrown error, and the fixture is quietly degraded for every other reader.

Nothing here is broken in CI — CI is green, the committed golden is intact, and the failure only appears once someone regenerates. It is a trap laid for the next person who does.

### Why the table is derivable

The hash is just the MD5 of the guidance bytes, because in the no-host-prefix baseline each guidance section *is* the whole system array. Verified against the committed table — all four rows reproduce byte-for-byte:

| Variant | Guidance bytes | MD5 | matches committed |
|---|---:|---|---|
| PRIMARY full | 8778 | `e8606e8e36cf19f5588ae0a76ce23b0e` | ✓ |
| PRIMARY memory-off | 7737 | `4d528493cac10523f815689677c46ea1` | ✓ |
| PRIMARY reduce-off | 6471 | `f5e8089b0ec037f8a81c648edd141913` | ✓ |
| SUBAGENT minimal | 691 | `83d69748f204fff98249565d5a31aa99` | ✓ |

So the generator can own the section outright and the table stops being hand-maintained.

### Fix

In `export-agent-surface.ts`, after the section-2 loop and before `outPath` is resolved, emit section 3 from the same `variants` array already in scope — `Buffer.byteLength(text, "utf8")` for the bytes column and `createHash("md5").update(text).digest("hex")` for the hash (plus `import { createHash } from "node:crypto"`). Table labels are the short form (`"PRIMARY full"`), i.e. `label.split(" (")[0]`.

I have this running on a downstream fork; regenerating there now reproduces upstream's committed section 3 exactly. Happy to open a PR if useful.

### Optional hardening

Independent of the fix, the two `slice(..., indexOf(...))` call sites silently accept `-1`. A guard that throws when the delimiter is missing would have surfaced this immediately instead of degrading to an empty tool set.

iceteaSA:
Closes #290. Both items you asked for, in one commit off clean `master` @ `77c51da9`.

## 1. The generator owns section 3

`export-agent-surface.ts` emitted only sections 1 and 2, so the hash baseline table was hand-maintained and regenerating the golden deleted it.

The table is now derived rather than transcribed: `Buffer.byteLength(text, "utf8")` and `createHash("md5").update(text)` over the same `variants` array section 1 already walks.

Verified by regenerating over the committed golden — output is byte-identical apart from the date stamp, all four hash rows included. One detail worth flagging: the committed prose contains `output.system.join("\\n")` (escaped backslash-n, not a newline). My first pass emitted `\n` and the diff caught it; the emitted string now reproduces the committed bytes exactly. The golden itself is untouched in this PR — the fix is the generator, and a date-only churn would just be noise.

## 2. Both slice call sites now throw

`tool-registry.test.ts` and `pi-plugin/src/tools/index.test.ts` both did:

```ts
document.slice(indexOf("## 2. Tool surface"), indexOf("## 3. System-prompt hash baseline"))
```

`-1` for a missing heading, so the slice returned `""`, the golden parsed as zero tools, and the tests **passed** comparing empty to empty. Agreed this is the half that hides the next regression. Both resolve offsets through a `sectionOffset()` helper that throws a named error.

### Red-check

With section 3 stripped from the golden:

```
error: A1 golden is missing the "## 3. System-prompt hash baseline" section heading
(fail) createToolRegistry — prompt-surface registration > matches the A1 golden …
(fail) registerMagicContextTools — prompt-surface registration > matches the A1 golden …
```

Same mutation on `master`: both green. That gap is the bug, and it is now closed on both hosts.

## Gates

| | |
|---|---|
| plugin | 3609 pass / 3 fail † |
| pi-plugin | 740 / 0 |
| cli | 296 (2 skip) |
| typecheck | 0 across 3 packages |
| lint | clean |

† Pre-existing `@opentui` TDZ errors in `tui-compiled-runtime-imports.test.ts` (`Cannot access 'TreeSitterClient' before initialization`). Unrelated to this change — it touches no TUI file — and reproducible on a clean `upstream/master` worktree with a fresh install. Your CI is green on master, so this looks local to my environment rather than something you need to act on; mentioning it only so the count is not mistaken for a regression here.

Files: `export-agent-surface.ts`, `tool-registry.test.ts`, `pi-plugin/src/tools/index.test.ts`.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/292"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788905200&installation_model_id=22398&pr_number=292&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F292&signature=ea77ebde34a0f3b2787308a876f2b85eb8bb895cc46fd725d99db69e4ca6ece2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Owns the A1 golden’s “System-prompt hash baseline” in the generator and makes tests fail when the section is missing. Prevents silent golden truncation and false green runs; also centralizes golden accessors to remove duplicated parsing and path math.

- **Bug Fixes**
  - `packages/plugin/scripts/export-agent-surface.ts` now emits section 3 by deriving the table from the guidance text (bytes via `Buffer.byteLength(..., "utf8")`, MD5 via `createHash("md5")`). Output matches the current golden (date-only diff).
  - `packages/plugin/src/plugin/tool-registry.test.ts` and `packages/pi-plugin/src/tools/index.test.ts` now throw when required section headings are missing, replacing the silent `slice(...)` behavior that produced empty tool sets.

- **Refactors**
  - Added `packages/plugin/src/shared/prompt-surface-a1-golden.ts` exporting `A1_TOOL_SECTION_HEADING`, `A1_HASH_BASELINE_HEADING`, `readA1GoldenDocument`, and `a1GoldenSectionOffset`. Both test suites now import these (Pi via `@magic-context/core/shared/prompt-surface-a1-golden`) to dedupe logic and ensure consistent guards.

<sup>Written for commit ceb4cb01ecacc1faa8062d32e9b29eb3770ef740. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/292?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

The PR makes the prompt-surface generator produce the system-prompt hash baseline and prevents malformed golden sections from silently yielding empty comparisons.
- Derives guidance byte counts and MD5 hashes from the generator’s existing variants.
- Centralizes golden-file loading, section headings, and missing-heading validation.
- Updates both plugin hosts’ tests to use the guarded shared helpers.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/plugin/scripts/export-agent-surface.ts | Generates section 3 from the same guidance variants used by section 1, including UTF-8 byte counts and MD5 hashes. |
| packages/plugin/src/shared/prompt-surface-a1-golden.ts | Adds shared golden-file access and explicit errors for missing required section headings. |
| packages/plugin/src/plugin/tool-registry.test.ts | Replaces unchecked heading offsets and local file-path handling with the guarded shared helpers. |
| packages/pi-plugin/src/tools/index.test.ts | Uses the shared golden reader and guarded section offsets while retaining the repository’s established cross-package import pattern. |
| packages/plugin/src/shared/prompt-surface-a1-golden.md | Contains the generated A1 prompt-surface snapshot consumed by both test hosts. |

</details>

<sub>Reviews (2): Last reviewed commit: ["refactor(prompt-surface): extract the A1..."](https://github.com/cortexkit/magic-context/commit/ceb4cb01ecacc1faa8062d32e9b29eb3770ef740) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51648203)</sub>


iceteaSA:
@cubic-dev-ai Valid — fixed in `ceb4cb01`.

New `packages/plugin/src/shared/prompt-surface-a1-golden.ts` owns the two heading constants, the document read, and the offset guard. Pi already consumes `plugin/src` through the `@magic-context/core/*` tsconfig alias, so this uses the established seam rather than introducing a new dependency.

One thing your report understated, and it is the better half of the fix: the duplicated **path arithmetic** was the more fragile part. The two readers resolved the same file as `../shared/…` and `../../../plugin/src/shared/…` respectively — invisible to the type checker, silently wrong the moment either file moves. The shared module resolves the golden relative to itself, so neither consumer carries a path any more.

Deliberately left out of scope: the two `readA1PrimaryGuidance()` readers in `system-prompt-hash.test.ts` and `pi-plugin/src/system-prompt.test.ts`. They regex the whole document instead of slicing between headings, so they were never exposed to the `-1` bug this PR fixes; folding them in would widen the PR into unrelated territory. The heading constants are exported and ready if a later change wants them.

Re-ran the red-check through the shared helper — with section 3 stripped from the golden, both consumers still fail with the named error:

```
error: A1 golden is missing the "## 3. System-prompt hash baseline" section heading
(fail) createToolRegistry — prompt-surface registration > matches the A1 golden …
(fail) registerMagicContextTools — prompt-surface registration > matches the A1 golden …
```

Gates unchanged: plugin 3609/3, pi 740/0, cli 296, typecheck 0 ×3, lint clean. The 3 are the pre-existing `@opentui` TDZ failures noted in the PR description.

Issue #312, opened 2026-08-13
Plugin tests share one context.db, and PRAGMA busy_timeout equals bun's default test timeout — lock contention surfaces as an unattributable 5000ms timeout


iceteaSA:
## What happens

`Check (plugin)` intermittently fails with tests that report `this test timed out after 5000ms` and nothing else. No assertion failure, no error message, no stack — just a wall-clock kill. The set of tests that fail changes between runs, and they pass on a re-run and in isolation.

Recent example on a PR branch: 3929 pass / 7 fail, all 7 timeouts, spread across four unrelated files (`system-prompt-hash`, `inject-compartments`, `compaction-off-transition`, `storage-embedding-measurements`). All four pass locally.

I originally read this as a slow runner, and want to show why that's wrong, because it's the reason the real cause is easy to miss:

| test | local | CI |
|---|---|---|
| `keeps project docs, user profile, and key files out…` | 52 ms | >5000 ms |
| `does NOT update systemPromptHash…` | 47 ms | >5000 ms |
| `injects the MINIMAL block for a subagent…` | 45 ms | >5000 ms |
| `invalidates cached marker reads after background writes…` | 242 ms | >5000 ms |
| `bounds a session's corpus rows…` | 133 ms | >5000 ms |

The CI runner is about 6.5× slower on this suite overall (668s vs 102s wall for the same 3926 tests). 6.5× on a 52 ms test is ~340 ms. These are 20–100× blowups. Nothing is *slow*; something is *waiting*.

## Cause

Three things upstream combine, and each is individually reasonable:

1. **`packages/plugin/test-preload.ts:30` points every test at one database.** It sets `MAGIC_CONTEXT_TEST_DATA_DIR` to a single temp dir, and `resolveDatabasePath()` (`storage-db.ts:162-172`) resolves a bare `openDatabase()` inside it. So every `openDatabase()` with no explicit path opens *the same* `context.db`. There are **491 such call sites across 54 test files**.

2. **`storage-db.ts:711` sets `PRAGMA busy_timeout=5000`.** Correct for production — the comment explains it guards a cold-open race between OpenCode/Pi startup and the subprocess lease tests.

3. **`bun test`'s default per-test timeout is also 5000ms**, and `packages/plugin/package.json`'s test script is a bare `bun test` with no `--timeout`.

All test files execute in **one process** (verified: two files report the same `process.pid`), so these connections are genuinely concurrent against one file.

[...]

PR #338, opened 2026-08-19
test: remove shared-DB and timer flakiness

### LINKED ISSUE #330: `storage-embedding-measurements` cap test does 2005 disk-backed inserts with no declared timeout — fails on slower hardware
## Summary

`embedding measurement corpus > bounds a session's corpus rows, keeping the newest when the cap is exceeded` times out at bun's default 5000 ms on my machine. It reproduces on **clean `upstream/master`**, in **isolation**, so it is not a fork regression and not the cross-file contention from #312.

```
$ cd packages/plugin && bun test src/features/magic-context/storage-embedding-measurements.test.ts
 1 pass
 1 fail          ← ^ this test timed out after 5000ms
Ran 2 tests across 1 file. [7.91s]
```

Verified on a pristine `git archive upstream/master` extraction with a fresh `bun install --frozen-lockfile`. Our fork's tree gives 7.85 s against upstream's 7.91 s — same failure, same timing, so the fork is not implicated.

## Cause

`storage-embedding-measurements.test.ts:58-88` loops `MEASUREMENT_CORPUS_SESSION_ROW_CAP + 5` times — the constant is **2000**, so **2005 iterations** — each calling `recordEmbeddingMeasurement` against a real on-disk SQLite database (`XDG_DATA_HOME` pointed at an `mkdtempSync` dir, opened via `openDatabase()`).

Measured cost is ~3.7 ms per insert here, so ~7.4 s of real work against a 5000 ms default. The test declares no timeout, so its verdict is a function of machine speed rather than of behavior.

Worth noting as a repo-wide observation rather than a one-test issue: `grep -c '}, [0-9]\{4,\});' src/**/*.test.ts` returns **0** — no test in the plugin package declares an explicit timeout, and `bunfig.toml` sets none. So every long test inherits 5000 ms silently. This one is simply the first to cross it.

## Why it isn't #312

[#312](https://github.com/cortexkit/magic-context/issues/312) is cross-file lock contention: many test files share one `context.db`, `PRAGMA busy_timeout` is 5000 ms, and bun's default per-test timeout is also 5000 ms, so a lock wait and the test kill expire together. That mechanism requires concurrency.

This one reproduces with a single file, a single test, and its own temp `XDG_DATA_HOME` — no second writer, no lock wait. The cause here is raw work volume against an undeclared timeout. Same symptom (5000 ms), unrelated mechanism.

## Suggested fix

Either is cheap; the first is narrower:

1. **Declare a timeout on this test** — `it("bounds a session's corpus rows…", () => { … }, 30_000)`. Honest about the work it does, and makes the verdict hardware-independent.
2. **Make the cap injectable for tests** so the loop can exercise the eviction boundary with, say, 25 rows instead of 2005. The test is verifying "keeps the newest when the cap is exceeded", which is a boundary property — it does not need the production cap value to prove it.

(2) is the better long-term shape: it tests the same invariant in ~90 ms instead of ~7.4 s, and removes the hardware dependence entirely rather than raising the ceiling above it.

## Environment

- Reproduced on clean `upstream/master` (v0.38.0) and on our fork, both with `bun install --frozen-lockfile`
- Linux, bun 1.3.14
- Passes in your CI, so this is hardware-dependent by nature — which is the point: the test's outcome should not depend on that

### LINKED ISSUE #333: `message-index-async` boot-quiet test has an 80ms wall-clock budget for two full reconcile cycles — fails under CI load
## Summary

`message-index-async > rebuilds when removal overtakes a boot-quiet reconciliation` failed a `Check (plugin)` run with an assertion failure at 114 ms (not a timeout). The test's verdict depends on wall-clock timing under load rather than on behavior.

Second instance of the class in [#330](https://github.com/cortexkit/magic-context/issues/330), different mechanism — #330 is raw work volume against the default timeout, this is a race window that must close inside a fixed `await`.

## What I can and cannot show

Being explicit about the evidence, because it is one observation:

- **Observed:** one failure in CI (run `32074486840`, job `95524587443`) on a fork branch based on `b27a6d76` (v0.38.0). `4118 pass / 1 fail`.
- **Not reproduced locally:** 14/14 in isolation, and **0 failures in 10 consecutive idle runs** on my machine.
- **The file is byte-identical to upstream** (`git diff --quiet upstream/master` on that path is clean), so the branch does not modify it.
- I could not re-run the job to check recurrence — no admin rights on the repo. Filing without that evidence rather than sitting on it.

So: **not proven systematic.** The structural argument below is why I think it is worth a look regardless.

## Mechanism

`message-index-async.test.ts:361-380`:

```ts
setBootQuietPeriodForTests(Date.now() - BOOT_QUIET_MS + 20);   // quiet expires in ~20ms

scheduleReconciliation(db, sessionId, readSurviving);
scheduleClearAndReindex(db, sessionId, readSurviving);

await wait(80);

expect(reads).toBe(2);
expect(countMessageRows(db, sessionId, "m-survivor")).toBe(1);
expect(isSessionReconciled(sessionId)).toBe(true);
```

What has to fit inside that 80 ms, twice over:

1. `scheduleAfterBootQuiet` → `setTimeout(task, bootQuietRemainingMs())` ≈ **20 ms** (`boot-quiet.ts:22`)
2. `defer(() => …)` — a second scheduling hop (`message-index-async.ts:208`)
3. `reconcileSessionIndex` / clear-and-reindex — **real SQLite work**, async
4. the `.finally()` bookkeeping that flips `isSessionReconciled`

So ~60 ms of slack covers two complete reconcile cycles including DB writes. A late `setTimeout` fire or slower disk on a shared runner eats that.

For contrast, the same file uses `await wait(140)` 10 times and `await wait(20)` 14 times — the 20 ms ones cover a single already-past-quiet defer, and the 140 ms ones have real slack. This is the only site where a ~20 ms timer, two scheduling hops, and two DB cycles all have to land inside one fixed window.

## One cross-signal worth noting

On my hardware, this test passes and **#330's** `storage-embedding-measurements` cap test fails. On the CI runner, the reverse: #330's test passes (faster box) and this one fails. Two different tests, two different machines, each failing on the machine that happens to disfavor it.

That is the general shape of the problem rather than an argument about either test specifically: the plugin suite currently has at least two verdicts that are functions of hardware speed. Related to my note in #330 that **no test in the package declares an explicit timeout** — the defaults are doing load-bearing work they were not chosen for.

## Suggested fix

Prefer determinism over a bigger number:

1. **Drive the scheduling with fake timers** so the test asserts ordering rather than racing a stopwatch. `setBootQuietPeriodForTests` already exists as a seam and its own docstring says "test seam for deterministic fake-timer coverage" — this test uses it with real timers.
2. Failing that, **poll for the condition** instead of sleeping a fixed budget (`await until(() => reads === 2, { timeout: 5_000 })`), so a slow runner is slow rather than red.

Raising 80 → 300 would also work but only moves the threshold; the verdict stays hardware-dependent.

## Environment

- Failure: GitHub Actions `ubuntu-latest`, run `32074486840`
- Local non-repro: Linux, bun 1.3.14, 16 cores idle
- Upstream `master` CI is green at `b27a6d76`, so this is not a standing breakage

coleleavitt:
## Summary
- set plugin test timeout to 30s so SQLite's 5s busy-timeout reports the actual lock stack instead of Bun killing the test first
- batch the 2,005 embedding-measurement setup inserts in one SQLite transaction
- return real completion promises from message-index schedulers and await them with a fake boot-quiet clock
- warm `@opentui/core` before parallel runtime-specifier imports to remove the `TreeSitterClient` TDZ race exposed by CI

Addresses #312 (improves lock-failure observability; per-file DB isolation remains open)
Fixes #330
Fixes #333

## Verification
- targeted storage/scheduler scope: 16 pass
- two concurrent targeted runners: both exit 0
- 20 repeated scheduler runs: pass
- 20 repeated TUI runtime-import runs: pass
- `bun run lint`
- `bun run typecheck`
- plugin build

Assertions are unchanged; no SQLite timeout was hidden or extended.

The PR reduces SQLite and scheduler test flakiness without changing production database timeout behavior.
- Extends the plugin test timeout to 30 seconds so SQLite lock failures surface before the test runner terminates them.
- Batches embedding-measurement fixture inserts in one transaction.
- Exposes scheduler completion promises and uses a fake boot-quiet clock in the overlap test.
- Serializes the TUI core/testing module initialization path.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/plugin/package.json | Raises the package test timeout so SQLite's configured busy timeout can produce a useful failure first. |
| packages/plugin/src/features/magic-context/message-index-async.ts | Returns deduplicated completion promises from reconciliation and clear/reindex scheduling while preserving per-session serialization. |
| packages/plugin/src/features/magic-context/message-index-async.test.ts | Replaces wall-clock waiting with fake timers and awaits both scheduler completion promises. |
| packages/plugin/src/features/magic-context/storage-embedding-measurements.test.ts | Batches the large measurement-corpus fixture setup in one SQLite transaction. |
| packages/plugin/src/tui/tui-compiled-runtime-imports.test.ts | Initializes the TUI core module before parallel runtime imports to avoid a module-initialization TDZ. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Test
    participant BootQuiet as Boot-quiet scheduler
    participant Lock as Session lock
    participant Index as Message index
    Test->>BootQuiet: schedule reconciliation
    Test->>BootQuiet: schedule clear and reindex
    Test->>BootQuiet: advance fake clock
    BootQuiet->>Lock: queue reconciliation
    BootQuiet->>Lock: queue clear and reindex
    Lock->>Index: reconcile existing messages
    Lock->>Index: clear indexed session
    Lock->>Index: rebuild from surviving messages
    Index-->>Test: resolve completion promises
```
</details>

<sub>Reviews (3): Last reviewed commit: ["test(tui): stabilize runtime import orde..."](https://github.com/cortexkit/magic-context/commit/d166e56b1016b7cddaa50618a30639970fa536a6) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54671383)</sub>

**Context used:**

- Knowledge Base — [Plugin memory, notes, and context tools](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/plugin-memory-and-tools.md)
- Knowledge Base — [Magic Context installation and release flow](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/installation-and-release.md)

### DIFF
--- a/packages/plugin/package.json
+++ b/packages/plugin/package.json
@@ -39,7 +39,7 @@
     "build:tui": "bun scripts/build-tui.ts",
     "check:tui-compiled": "bun run build:tui && git diff --exit-code -- src/tui-compiled && test -z \"$(git status --porcelain -- src/tui-compiled)\"",
     "typecheck": "tsc -p ../retina-local-fs/tsconfig.build.json && tsc --noEmit && tsc -p tsconfig.scripts.json",
-    "test": "bun test --isolate",
+    "test": "bun test --timeout 30000",
     "lint": "biome check .",
     "lint:fix": "biome check --write .",
     "format": "biome format --write .",
--- a/packages/plugin/package.json
+++ b/packages/plugin/package.json
@@ -39,7 +39,7 @@
     "build:tui": "bun scripts/build-tui.ts",
     "check:tui-compiled": "bun run build:tui && git diff --exit-code -- src/tui-compiled && test -z \"$(git status --porcelain -- src/tui-compiled)\"",
     "typecheck": "tsc -p ../retina-local-fs/tsconfig.build.json && tsc --noEmit && tsc -p tsconfig.scripts.json",
-    "test": "bun test",
+    "test": "bun test --isolate",
     "lint": "biome check .",
     "lint:fix": "biome check --write .",
     "format": "biome format --write .",
--- a/packages/plugin/src/features/magic-context/message-index-async.test.ts
+++ b/packages/plugin/src/features/magic-context/message-index-async.test.ts
@@ -1,6 +1,6 @@
 /// <reference types="bun-types" />
 
-import { afterEach, beforeEach, describe, expect, it } from "bun:test";
+import { afterEach, beforeEach, describe, expect, it, jest } from "bun:test";
 
 import type { RawMessage } from "../../hooks/magic-context/read-session-raw";
 import { BOOT_QUIET_MS, setBootQuietPeriodForTests } from "../../plugin/boot-quiet";
@@ -114,6 +114,7 @@ describe("message-index-async", () => {
 
     afterEach(() => {
         setBootQuietPeriodForTests(null);
+        jest.useRealTimers();
         closeQuietly(db);
         __resetMessageIndexAsyncForTests();
     });
@@ -359,6 +360,8 @@ describe("message-index-async", () => {
     });
 
     it("rebuilds when removal overtakes a boot-quiet reconciliation", async () => {
+        jest.useFakeTimers();
+        jest.setSystemTime(0);
         const sessionId = "ses-boot-clear";
         const surviving = [message("m-survivor", 1, "surviving searchable bytes")];
         let reads = 0;
@@ -366,13 +369,15 @@ describe("message-index-async", () => {
             reads++;
             return surviving;
         };
-        setBootQuietPeriodForTests(Date.now() - BOOT_QUIET_MS + 20);
+        setBootQuietPeriodForTests(0);
 
-        scheduleReconciliation(db, sessionId, readSurviving);
-        scheduleClearAndReindex(db, sessionId, readSurviving);
+        const reconciliationFinished = scheduleReconciliation(db, sessionId, readSurviving);
+        const rebuildFinished = scheduleClearAndReindex(db, sessionId, readSurviving);
         expect(isSessionReconciled(sessionId)).toBe(false);
 
-        await wait(80);
+        jest.advanceTimersByTime(BOOT_QUIET_MS);
+        jest.runAllTimers();
+        await Promise.all([reconciliationFinished, rebuildFinished]);
 
         expect(reads).toBe(2);
         expect(countMessageRows(db, sessionId, "m-survivor")).toBe(1);
--- a/packages/plugin/src/features/magic-context/message-index-async.ts
+++ b/packages/plugin/src/features/magic-context/message-index-async.ts
@@ -73,7 +73,7 @@ const INCREMENTAL_DEBOUNCE_MS = 100;
 const RECONCILIATION_BATCH_SIZE = 100;
 
 const reconciledSessions = new Set<string>();
-const reconciliationScheduledSessions = new Set<string>();
+const reconciliationScheduledSessions = new Map<string, Promise<void>>();
 const sessionLocks = new Map<string, Promise<void>>();
 const incrementalTimers = new Map<string, ReturnType<typeof setTimeout>>();
 const pendingIncrementalKeys = new Set<string>();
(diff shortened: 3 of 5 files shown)


alfonso-magic-context:
Thank you, Cole — filed before our independent #330/#333 changes landed, and your read of the timing signature was useful, especially the #312 point that Bun's 5s test timeout and SQLite's 5s busy_timeout hide lock errors behind the same wall-clock cutoff.

In parallel, master landed different implementations: #330 now exercises the corpus-eviction invariant with an injectable small cap instead of 2,005 disk-backed inserts, and #333 waits for the observable rebuild state instead of a fixed sleep — so we don't want to merge a second implementation of those two.

Two pieces of this PR are independently useful and we're extracting them with credit: `bun test --timeout 30000` so a SQLite lock surfaces as a lock error before the runner kills the test, and the `@opentui/core` warm-up before parallel runtime-specifier imports (a real TDZ race our suite hasn't fixed). To be explicit: the timeout change is diagnostic — it doesn't isolate per-file databases or change test busy_timeout, so #312 stays open for the real contention fix. Closing this PR as superseded; the extractions will land crediting you.

Issue #334, opened 2026-08-18
`MemorySourceType` declares a `"user"` variant that nothing ever writes — operator-stated facts are indistinguishable from agent inference


iceteaSA:
## Summary

`MemorySourceType` has four variants and the dashboard renders all four, but **only three are reachable**. Nothing in the codebase ever writes `sourceType: "user"`, so a fact the operator stated is stored byte-identically to one an agent inferred on its own.

Small and self-contained. Filing separately because it stands on its own regardless of the larger proposal that motivated me to notice it (see #335).

## Evidence

All references against `upstream/master` @ `b27a6d76` (v0.38.0).

The type declares it — `packages/plugin/src/features/magic-context/memory/types.ts:25`:

```ts
export type MemorySourceType = "historian" | "agent" | "dreamer" | "user";
```

The dashboard already renders it — `packages/dashboard/src/components/MemoryBrowser/MemoryBrowser.tsx:342-350`:

```ts
case "historian": return "purple";
case "agent":     return "blue";
case "dreamer":   return "indigo";
case "user":      return "green";     // ← unreachable
```

But every write site is one of the other three:

```
memory/promotion.ts:75                    sourceType: "historian"
memory/memory-migration.ts:233            sourceType: "historian"
dreamer/retrospective-learnings.ts:190    sourceType: "dreamer"
tools/ctx-memory/tools.ts:291             return deps.sourceType ?? "agent"
```

A repo-wide search for `sourceType: "user"` across non-test `.ts` on master returns nothing. On my store (1,431 active memories) the live distribution is `agent` 677 / `historian` 467 / `dreamer` 287 / **`user` 0**.

## Why it matters

When the operator says "always use X" and the agent records it with `ctx_memory`, the row lands as `source_type='agent'` — the same value as a memory the agent minted from its own reasoning with no human input. Downstream that means:

- **The dashboard's green `user` badge can never appear**, so the operator cannot visually separate their own stated rules from agent inferences when auditing the store.
- **`curate` / `verify` cannot weight operator-stated facts differently.** An agent inference and an operator instruction are equally eligible for archival. Given that archiving an operator-stated rule is strictly worse than archiving an agent guess, there is currently no signal to protect the former.
- Any future work wanting "operator said so" as a trust input has no column to read.

## Scope

[...]

PR #340, opened 2026-08-19
feat(memory): preserve corroborating episode evidence

### LINKED ISSUE #335: Memory promotion is single-episode: `seen_count` counts phrasing repeats, not corroboration — proposal for an episode table + optional tiering
## Summary

A single historian extraction can mint a durable project memory. There is already a replication counter — `memories.seen_count` — but it increments on **exact normalized-hash match**, so it measures whether the historian phrased a fact identically twice, not whether two sessions independently corroborated it.

This proposes one small primitive (an episode table) that makes corroboration actually computable, plus two optional policies on top. **The primitive is the ask; the policies are a sketch you can take or leave.**

All references against `upstream/master` @ `b27a6d76` (v0.38.0). Measurements from my own store (1,432 active memories) — the shapes should reproduce anywhere, the exact percentages will not.

## The gap

`memory/promotion.ts:63-68`:

```ts
const existingMemory = getMemoryByHash(db, projectPath, fact.category, normalizedHash);

if (existingMemory) {
    updateMemorySeenCount(db, existingMemory.id);
    continue;
}
```

Two consequences:

1. **Hash equality is the corroboration test.** Two sessions that independently learn the same thing but phrase it differently produce two rows at `seen_count=1`, not one at `seen_count=2`. The counter tracks the historian's phrasing consistency.
2. **The episode set is discarded.** The bump records a count and `last_seen_at`; it does not record *which* session bumped it. `source_session_id` is singular — the first writer only (167 distinct values across 1,432 rows here). So "corroborated by N independent sessions" cannot be answered even approximately.

Live distribution:

```
seen_count = 1   1259
           = 2    127
           = 3     31
           = 4      8
           = 5      6

verification_status = 'verified'    78  (5.5%, despite verify running nightly)
```

## The primitive (the actual ask)

```sql
CREATE TABLE memory_episodes (
    memory_id  INTEGER NOT NULL,
    session_id TEXT    NOT NULL,
    seen_at    INTEGER NOT NULL,
    actor      TEXT,
    PRIMARY KEY (memory_id, session_id)
);
```

Append at the existing bump site and at insert. The `PRIMARY KEY (memory_id, session_id)` is doing real work: it counts a **session** once regardless of how many times extraction restates the fact within it, so the count cannot be inflated by extraction volume.

That alone makes "how many distinct sessions, on how many distinct days, corroborated this?" answerable — which is useful for `curate`, for the dashboard, and for anyone reasoning about which memories have earned their place, **independent of whether you adopt any tiering policy**. Cost is one migration and roughly two lines in `promoteSessionFactsDurable`.

## Optional policy A — two tiers

If you want it: new facts land episodic, promote to durable on N distinct sessions (proposed N=2 with day-independence; N=3 plus actor-independence for `CONSTRAINTS`/`PROJECT_RULES`, where a wrong memory costs a wrong unattended decision), or on an existing verification artifact. Demotion on contradiction goes to a visible `disputed` state rather than a silent overwrite. Promotion records which criterion fired.

**Two things worth knowing before you consider this:**

**The bar is brutal on a real corpus.** Under `seen_count >= 2 OR verified`, my store promotes **247 of 1,432** — so **1,185 (83%) would sit episodic**. Some of that is the hash-equality flaw above (real corroboration miscounted), but not all. The criteria and N are a dial and the episode table is what makes any setting measurable; I would not ship a default without measuring on your own corpus first.

**Quarantine should be non-injection, not a marker.** #320 established that a "this is data, not instructions" preamble on `<project-memory>` demotes `PROJECT_RULES`, which exist to bind behavior. So episodic facts should simply **not render** until promoted, rather than rendering with a trust marker. No new prompt wording to freeze forever.

**Cache note**, since this touches the injected set: a tier flip changes *which* memories render, so it is not cache-neutral the way `setMemoryClassification` deliberately is (`storage-memory.ts:1117-1120` documents that contrast). It has to ride the existing non-additive path — `memory_mutation_log` → `<memory-updates>` in m[1] → reconcile into m[0] on the next natural hard fold. The hazard is batch shape rather than mechanism: a nightly promotion burst writes many mutation rows at once, inflating m[1] in one step, which can trip the m[1] pressure-backstop refold across active sessions the following morning. Mitigations: cap promotions per night, or let them accumulate and flush on each session's next natural fold instead of minting deltas eagerly.

## Optional policy B — rank consolidation by surprise

Extraction depth currently does not track information content. In practice the memories that earn their keep disproportionately come from corrections, refutations, and gate failures after green claims, while routine-success episodes tend to restate what the docs already say.

The non-obvious part is **where this has to live**. The dreamer looks like the natural home, but its only transcript surface is user-only:

```
retrospective-raw-provider.ts:414   "the newest `count` USER rows. Returns user-only"
```

The historian reads assistant, user, and tool rows (`read-session-db.ts:108/145/247`). So of the marker classes worth detecting — operator corrections, REFUTED/verdict reversals, explicit retraction language, self-caught errors — **the dreamer can see only the first**. Detection belongs in the historian, which already has the bytes in hand at extraction time; the dreamer's batch pass consumes the flags to allocate depth. Nearly free, no second transcript read.

Guards if pursued: small integer weight tiers (not learned weights), artifact markers (gate output, review verdicts) outrank self-reported ones, density caps so repeated "correction" language cannot farm rank — and surprise **ranks but never filters**, so a zero-marker episode holding a first-seen fact still gets extracted.

Episode-dedup is what keeps A and B independent: deeper extraction of a surprising episode must not inflate promotion counts, and `PRIMARY KEY (memory_id, session_id)` guarantees it cannot.

## What I am asking

Interest in the **episode table** specifically. It is small, useful on its own, and no policy has to follow it. If you want it, I will send a PR for just that — migration, the two write-site lines, tests — and we can discuss tiering separately or not at all.

Prerequisite for the "operator ratified it" criterion is filed separately as #334 (the `"user"` source-type variant is declared and rendered but never written).

Design provenance: this came out of a cross-project design exchange; I attacked both drafts against this codebase and the findings above are what survived. Cross-links: #334, #320.

coleleavitt:
## Summary
- add content-versioned per-session evidence rows for TypeScript and Rust memory stores
- make exact repeats idempotent within one session while distinct sessions increase corroboration
- preserve evidence through update, merge, move/copy collisions, v22 backfill, identity merge, Pi merge, deletion, and TS↔Rust authority transfer
- keep `seen_count` monotonic and based on distinct corroborating sessions without reducing legacy aggregate counts
- document the cross-runtime source contract and migration v79/v51

Fixes #335
Related to #334

## Attribution boundary
OpenCode supplies the assistant message that owns the tool call, not a verified originating user message. Primary writes therefore remain `source_type=agent`; the dormant `user` variant is not falsely populated. `source_message_id` is the host-native assistant tool owner on OpenCode and null on Pi/Rust. A future verified host user-message link can activate `user` safely.

## Semantic boundary
Evidence records observations of exact content versions. It does not claim paraphrased memories are the same assertion; explicit merge retains the original hashes.

## Verification
- plugin provenance/migration/authority/identity/backfill/tool suites: 121 pass
- complete Pi ctx_memory suite, including restored pre-existing coverage: 25 pass
- plugin and Pi typecheck/build pass
- plugin lint passes; Pi has one pre-existing non-null warning
- Rust `mc-store` compiled against public `cortexkit\/commons`; 136\/136 store tests pass. Full `mc-module` workspace tests remain blocked by unavailable `subconscious` path dependencies
- three Oracle reviews found and verified fixes for false user attribution, lifecycle evidence loss, non-atomic writes, content-version drift, distinct-session counting, legacy count monotonicity, and TS/Rust merge parity

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/340"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789717932&installation_model_id=22398&pr_number=340&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F340&signature=34f4846eec29d2f8b929723289974754e7cfdeae5904b9b9d5607be55f504218"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Preserves content-bound, per-session episode evidence for memories and makes duplicate writes idempotent within a session. Old: duplicate writes bumped seen_count without provenance and evidence-only updates sometimes didn’t sync; new: evidence keyed by content hash and session recomputes seen_count from distinct sessions, and evidence-only changes now sync and seed mirrors in all runtimes.

- Review and migration
  - Schema v79: create `memory_evidence` (PK: `memory_id, content_hash, source_session_id`) with a session index; backfill from `memories`; bump `LATEST_SUPPORTED_VERSION` to 79; replay-safe for armed stores.
  - Behavior: use `insertMemoryIdempotent` and `recordMemoryEvidence` instead of ad‑hoc seen_count bumps; same-session repeats are no-ops; distinct sessions add corroboration. Attribution requires a verified owner for `source_message_id`; tool writes default `source_type=agent` and never infer `user`.
  - Lifecycle and counts: updates, rekeys/collision merges, identity merges, v22 backfill, copies, and deletes union or cascade evidence; when evidence exists, seen_count derives from distinct-session evidence, otherwise the legacy max is preserved.
  - Cross‑runtime: state‑sync payloads and authority seeds carry evidence arrays; mirror sync reinstalls evidence snapshots; evidence‑only updates rewind watermarks to force resend; TypeScript and Rust stores persist/union evidence and maintain distinct‑session counts.

<sup>Written for commit 1975b8a81574d0881b88d3a0291016e3a7911510. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/340?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

The PR adds content-versioned, per-session evidence for durable memories across the TypeScript, Pi, and Rust stores.

- Adds evidence schemas and migrations for both storage implementations.
- Preserves and unions evidence across memory lifecycle operations and authority transfers.
- Makes same-session repeats idempotent while distinct sessions increase corroboration.
- Updates state synchronization to include evidence and resend evidence-only changes to existing memory rows.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/plugin/src/tools/ctx-memory/tools.ts | Duplicate writes now request Rust synchronization before returning, covering both existing-row and unique-constraint-race outcomes. |
| packages/plugin/src/features/magic-context/memory/storage-memory.ts | Adds idempotent per-session evidence recording and monotonic distinct-session corroboration accounting. |
| packages/plugin/src/hooks/magic-context/module-state-sync.ts | Includes memory evidence in cross-runtime state payloads so synchronized rows preserve provenance. |
| packages/plugin/src/hooks/magic-context/rust-mode-transform.ts | Rewinds memory-specific watermarks when a tool mutation requests synchronization, ensuring evidence-only changes to existing IDs are resent. |
| crates/mc-store/src/lib.rs | Adds the Rust evidence schema and lifecycle handling needed for TypeScript-to-Rust parity. |
| packages/pi-plugin/src/tools/ctx-memory.ts | Adopts the shared idempotent memory insertion and evidence semantics for Pi writes. |

</details>

```mermaid
flowchart LR
  W[Memory write or repeat] --> TS[(TypeScript memory store)]
  TS --> E[(Per-session evidence)]
  E --> C[Recompute monotonic seen_count]
  E --> Q[Request Rust memory sync]
  Q --> R[Rewind memory watermarks]
  R --> P[Build state-sync payload]
  P --> RS[(Rust memory store)]
  RS --> M[Mirror or authority transfer]
  M --> TS
```
</details>

<sub>Reviews (7): Last reviewed commit: ["style(pi): format memory evidence regres..."](https://github.com/cortexkit/magic-context/commit/1975b8a81574d0881b88d3a0291016e3a7911510) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=54679490)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Plugin memory, notes, and context tools](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/plugin-memory-and-tools.md)
- Knowledge Base — [Rust memory and context engine](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/rust-memory-engine.md)
- Knowledge Base — [Pi plugin runtime: adapting Magic Context to Pi](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/pi-plugin-runtime.md)
</details>


coleleavitt:
Retriggering the immutable head because the only failed check was a pre-code OpenCode version-download failure in the non-required 0.0.0.0 smoke probe.

iceteaSA:
Ran two independent cross-family reviews of this PR against the design in #335, then verified every finding below myself against the PR branch (`feat/memory-episode-provenance` @ `1975b8a8`). Reporting because we filed the issue — take or leave any of it.

**First: the invariant we cared about holds, and our prescription was wrong.**

#335 argued for `PRIMARY KEY (memory_id, session_id)` specifically so one session couldn't inflate a corroboration count by restating a fact N ways. You used a 3-part key with `content_hash` and enforced the invariant in the *counting* instead:

```
COUNT(DISTINCT source_session_id)   × 3 sites in storage-memory.ts
COUNT(*) over memory_evidence       × 0
```

That's strictly better than what we asked for — it preserves content-version provenance that our 2-part key would have destroyed, and `storage-memory-evidence.test.ts` pins the exact case (one session, two content versions → 2 evidence rows, `seenCount === 1`). Our issue was wrong on the mechanism; the reviewers both landed on "content_hash earns its place." Worth saying plainly.

Also confirmed clean: `LATEST_SUPPORTED_VERSION` = 79 in lockstep, `memory_evidence` created in both the v79 migration *and* the fresh-DB path, no new `mustMaterialize` trigger, additive memories still ride m[1].

## Two findings both reviewers reached independently

Both reduce to the same root: **the evidence table has no session-lifecycle story.** Flagging now rather than later because nothing gates promotion on evidence yet (`grep evidenceCount|corroborat` → 0 hits in `storage-memory.ts`; permanence still uses `retrieval_count`), so today this is bookkeeping drift — it becomes correctness the moment a threshold reads these rows.

**1. `clearSession()` doesn't reap evidence.** The PR has exactly one `DELETE FROM memory_evidence`, and it's `WHERE memory_id = ?` — memory-scoped, not session-scoped. `storage-meta-session.ts` isn't touched by the diff at all. A reviewer's in-memory probe after `clearSession(db, "ses-leak")` returned `{"evidence":1,"memories":1}`.

`ARCHITECTURE.md` (Storage & migrations) states the rule: *"New session-scoped tables must be added to `clearSession()`."* `memory_evidence` carries `source_session_id`, so it qualifies.

Genuinely two-sided though, and the second reviewer argued the other side well: deleting evidence means a memory silently *loses* corroboration it legitimately earned. Its proposed third option — roll a terminal session's distinct contribution into a compact per-memory count, then drop the detail rows — preserves the count without keeping dead session references forever. Your call which semantics you want; the current state looks like neither choice was made explicitly.

**2. No harness discriminator, which makes safe cleanup impossible.** `memory_evidence` is `(memory_id, content_hash, source_session_id, source_message_id, source_type, observed_at)`. No `harness` column — while `session_meta` has one.

This one I can speak to from operational experience rather than reading. Yesterday I found **1,992 of 6,443 `session_meta` rows orphaned** on my store (31%) — sessions deleted out-of-band without emitting `session.deleted`, so the event-driven cleanup never fired. Reaping them safely was only possible *because* `session_meta` has `harness`: **Pi sessions are legitimately absent from `opencode.db` by design**, so the obvious "delete rows with no matching opencode session" query destroys Pi metadata. I had exactly one such row and nearly took it.

`memory_evidence` can't be swept that way at all. There's an index on `(source_session_id, memory_id)` so the lookup is cheap — it's purely the missing discriminator. The reviewer also notes host-local session ids could collide across harnesses under the current PK.

## Three Shoulds, lower confidence — evidence attached, judge for yourself

- **Publish-transaction cost.** Each new corroboration adds two `COUNT(DISTINCT …)` scans plus insert/update/reload inside `promoteSessionFactsDurable`'s transaction; SQLite reports `USE TEMP B-TREE FOR count(DISTINCT)`. Given #312 (shared DB + `busy_timeout=5000`), lock-hold duration has bitten here before. A `(memory_id, source_session_id)` index would drop the temp b-tree.
- **N+1 in forced module sync.** `module-state-sync.ts` issued 1,443 queries for 1,443 memories on a production copy — measured 9.344 ms warmed vs 0.824 ms bulk-loaded.
- **Pi divergence undocumented.** OpenCode records `sourceMessageId`; Pi leaves it null. `MEMORY-DESIGN.md:170-177` explains why, but `PARITY.md` is where deliberate divergences are registered.

## Growth, for sizing

No retention, pruning, cap, or sweep. Rows grow as `Σ_memory distinct(content_hash, source_session_id)`. On my corpus (1,432 memories, 6,443 sessions) the v79 migration created 2,272 rows at ~232 bytes/row including indexes. The reachable bound at 3 content versions is 1,432 × 6,443 × 3 ≈ 27.7M rows ≈ 6.4 GB — worst-case, not expected-case, but there's currently nothing that makes it unreachable.

## What we could not check

Rust evidence tests didn't run — the cloned workspace was missing `commons/crates/cortexkit-cache-core`, so that leg is **unverified**, not passed.

What I did verify by reading the PR branch directly: `promote_facts_tx` in `crates/mc-store/src/lib.rs` contains **0** calls to `record_memory_evidence_tx` (the recorder is defined in the same file), and `to_store_fact` at `crates/mc-module/src/historian.rs:74-80` sets `source_session_id: None` in production. The TS counterpart passes `sourceSessionId: sessionId` (`memory/promotion.ts:60`). If that's deliberate sequencing for a later PR, ignore this — but as it stands the two runtimes disagree about whether historian promotion produces evidence at all.

Happy to send a PR for the `clearSession` reap and/or the harness column if either is wanted, or to leave it entirely to you.

alfonso-magic-context (changes requested):
Reviewed in depth — and worth saying first: you and @iceteaSA (#335) independently converged on the same missing primitive, which is strong signal the gap is real. The design direction is right: content-bound `(memory, content hash, session)` episodes make independent-session corroboration computable without adding a category, changing the uniqueness key, or touching rendered memory bytes.

Two blockers before this can merge:

1. **Rust-mode historian promotion records no episodes.** `to_store_fact()` leaves `source_session_id` as `None`, `publish_historian_chunk()` calls `promote_facts_tx()` without the publishing session, and the exact-duplicate skip path never records evidence. In rust transform mode the historian is the main autonomous memory producer, so as delivered the corroboration signal would silently not accumulate exactly where it matters most. The new Rust tests cover direct insert/facade writes but not historian publication — that path needs both the wiring and a test.
2. **`memory_evidence` needs to join the authority guard fence.** It mirrors from the module authority store, but the TypeScript side has no guard triggers for it, and `recordMemoryEvidence()` doesn't assert write permission itself — today it's protected only incidentally because episode writes also touch the guarded parent row. An evidence-only write path added later could mutate a MODULE-authority project's read model unguarded. It should get its own triggers plus coverage in the armed-replay test alongside the six existing guarded tables.

Change requests are narrow and the core design stands. If you'd rather we take the fence work (it's deep in our authority machinery), say so and we'll layer it as a follow-up commit on your PR.

alfonso-magic-context:
Design update that affects this PR: we've taken the corroboration question to a public design discussion in #335 (full reasoning there, including why session-keyed episodes conflict with the operating model we're converging on and what we're proposing instead — deterministic decay + a two-step historian observation-reconciliation). Parking this PR pending that thread's outcome rather than asking for the rework from the earlier review — your input on the #335 direction is very welcome, and the review's rust-wiring and authority-fence findings will apply to whatever variant emerges.

coleleavitt:
Closing recommendation: close without merge as superseded design work, while keeping #335 open. The content-versioned implementation established useful invariants and tests, but the #335 design discussion rejected session identity as the long-term corroboration unit for permanent-session operation. None of this PR's commits landed on master.

The branch is now also mechanically unsafe to rebase as-is: its TS migration v79 and Rust migration v51 have both been assigned to unrelated released schemas upstream. The requested Rust-historian publication and authority-fence fixes remain unresolved at this head, and any successor also needs explicit harness identity and lifecycle/retention semantics.

Current master contains neither this evidence table nor the proposed observation-reconciliation/decay replacement, so closing this PR must not close #335. If provenance returns, carry forward the content-hash/distinct-observation tests plus Rust historian, authority, harness identity, retention, and newly allocated migration coverage.

Issue #247, opened 2026-07-25
[Pi] In-process parallel subagents can load Magic Context repeatedly and OOM in SessionManager.listAll


beilo:
### Short description

In-process `@gotgenes/pi-subagents` children inherit and initialize Magic Context. Four parallel children can trigger concurrent global Pi session scans and crash the parent Node process with heap OOM.

### What happened?

Environment where this reproduced:

- Magic Context: `@cortexkit/pi-magic-context` 0.32.4
- Pi: `@earendil-works/pi-coding-agent` 0.81.1
- Subagents: `@gotgenes/pi-subagents` 18.1.0
- Node: 22.23.0
- Platform: macOS arm64
- Pi session store: about 392 JSONL sessions

Reproduction shape:

1. Register both `npm:@gotgenes/pi-subagents` and `npm:@cortexkit/pi-magic-context` in Pi user packages.
2. Start a normal persisted Pi session and let Magic Context finish its boot-quiet period.
3. Start four background `@gotgenes/pi-subagents` children in parallel. Each child only needs one MCP tool call.
4. The parent Pi process terminates with `SIGABRT` / JavaScript heap OOM.

Observed failure, sanitized:

```text
Mark-Compact 4010.4 (4352.5) -> 3255.4 (4333.2) MB ... allocation failure
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

node::OOMErrorHandler
v8::internal::V8::FatalProcessOutOfMemory
v8::internal::JsonParser<...>::ParseJson
Builtin_JsonParse
node::fs::FSReqCallback::Resolve
```

This reproduced repeatedly with the OOM stack in JSON parsing during an async filesystem callback. Running without Magic Context removes this parallel-start crash path; Pi native compaction remains usable.

### Why this appears different from #238

#238 covers Magic Context's own subprocess-based historian/dreamer children. Those children use `MAGIC_CONTEXT_PI_SUBAGENT=1`, and v0.33.0 adds isolated retry / extension controls for that runner.

`@gotgenes/pi-subagents` 18.x uses in-process `createAgentSession()` children instead:

- its child loader intentionally inherits all parent extensions;
- it emits `subagents:child:session-created` immediately before `bindExtensions()`;
- it does not set `MAGIC_CONTEXT_PI_SUBAGENT=1` because `process.env` is process-global and unsafe as a per-child marker.

[...]

Issue #177, opened 2026-06-20
[pi] Historian error spawning pi (Windows Powershell)


davidambler:
### Short description

When running Magic Context via the Pi plugin on Windows, the historian subagent fails to start with repeated spawn_failed errors: `Pi recomp historian failed (spawn_failed): spawn pi ENOENT`

### What happened?

When running Magic Context via the Pi plugin on Windows, the historian subagent fails to start with repeated spawn_failed errors: `Pi recomp historian failed (spawn_failed): spawn pi ENOENT`

## Environment
- Pi plugin: v0.26.0
- Pi: not installed
- OS: win32 x64
- Node: v26.2.0

## Diagnostics
- Timestamp: 2026-06-20T13:46:40.357Z
- Pi plugin: v0.26.0
- OS: win32 x64
- Node: v26.2.0
- Pi installed: true
- Magic Context package registered: true
- User config parse error: none
- Project config parse error: none
- Known Pi extension conflicts: none

### Pi settings
```json
{
  "path": "<HOME>\\.pi\\agent\\settings.json",
  "exists": true,
  "hasMagicContextPackage": true,
  "packages": [
    "npm:pi-mono-multi-edit",
    "npm:@hugo-hsi-dev/pi-interactive-questionnaire",
    "npm:pi-ocr",
    "npm:pi-lens",
    "npm:@tintinweb/pi-subagents",
    "npm:pi-powerline-footer",
    "npm:@cortexkit/pi-magic-context"
  ]
}
```

### Config paths
```json
{
  "agentDir": "<HOME>\\.pi\\agent",
  "userConfig": "<HOME>\\.pi\\agent\\magic-context.jsonc",
  "projectConfig": "D:\\project\\.pi\\magic-context.jsonc"
}
```

### User magic-context.jsonc flags
```jsonc
{
  "$schema": "https://raw.githubusercontent.com/cortexkit/magic-context/master/assets/magic-context.schema.json",
  "historian": {
    "model": "lm-studio/qwen3.6-27b-text-nvfp4-mtp"
  },
  "dreamer": {
    "model": "lm-studio/qwen3.6-27b-text-nvfp4-mtp"
  },
  "sidekick": {
    "model": "lm-studio/qwen3.6-27b-text-nvfp4-mtp"
  },
  "embedding": {
    "provider": "local",
    "model": "Xenova/all-MiniLM-L6-v2"
  }
}
```

### Project magic-context.jsonc flags
```jsonc
{}
```
[...]

PR #350, opened 2026-08-21
fix(pi-plugin): support pi-web multi-session and RPC hosts

elrond298:
## Summary

This PR adapts the Pi plugin for [`pi-web`](https://github.com/agegr/pi-web), where multiple Pi sessions share one persistent Node.js process and commands run through Pi’s RPC mode.

It addresses shared-process session isolation, duplicate startup work, unsafe subagent CLI detection, RPC command feedback, and `/ctx-dream` failures before the first model turn.

### Session isolation

- Replace the process-global child-session flag with an `AsyncLocalStorage<boolean>` marker.
- Update the marker through the public `subagents:child:session-created` and `subagents:child:disposed` events.
- Suppress initialization only for in-process child subagents.
- Allow independent sessions in the same `pi-web` process to initialize normally.
- Remove child lifecycle listeners during `session_shutdown` to prevent stale handlers after reloads.

### Process-wide startup maintenance

- Claim deferred startup maintenance once per process.
- Schedule the legacy-memory identity rekey introduced by database migration v22 only once per process.
- Schedule the session-to-project mapping backfill only once per process.
- Load session history only after acquiring the durable backfill lease, avoiding redundant JSONL scans.

### Safer subagent CLI detection

- Reuse `process.argv[1]` only when it identifies a supported Pi CLI; otherwise use the packaged executable, bundled CLI, or `PATH` fallback as appropriate.
- This prevents embedded hosts such as Next.js from accidentally launching another web server when creating a subagent.

### RPC command presentation

- Preserve model-invisible command entries while presenting them through the RPC UI.
- Show short progress updates through `ctx.ui.notify`.
- Show formatted results in modal dialogs through `ctx.ui.custom`.
- Apply this behavior to `/ctx-status`, `/ctx-embed`, `/ctx-recomp`, `/ctx-session-upgrade`, and `/ctx-dream`.

### Dreamer registration

- Track Dreamer registration ownership per Pi extension instance.
- Prevent one session’s shutdown from deregistering a same-project sibling.
- Transfer the active registration to a remaining worktree owner when necessary.
- Synchronize the current project’s Dreamer registration immediately before a manual `/ctx-dream` run.
- This allows `/ctx-dream` to work before the first `before_agent_start` event.

## Verification

- `bun run --cwd packages/pi-plugin build`
- `bun run --cwd packages/pi-plugin test` — **804 passed, 0 failed**
- `bun test packages/plugin/src/features/magic-context/session-project-backfill.test.ts` — **9 passed, 0 failed**
- `git diff --check`

Regression coverage includes:

- independent same-process sessions versus child-session suppression
- lifecycle listener cleanup during shutdown
- process-wide startup maintenance claimed through full runtime initialization
- lazy backfill lease gating without session-history reads
- owner-aware Dreamer registration, shutdown isolation, and ownership transfer
- single-timer Dreamer handoff when remaining owners repeat a worktree directory
- embedded-host Pi CLI discrimination
- RPC notification and modal-dialog routing
- pre-execution `/ctx-dream` registration synchronization

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/350"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789875783&installation_model_id=22398&pr_number=350&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F350&signature=51b0c05e565aaebe00749541adf8d34b70ef851c1716887bed3f75375e82a3ff"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Supports `packages/pi-plugin` in pi-web multi-session and RPC hosts. Old behavior suppressed all same-process inits and drained all jobs on shutdown; new behavior suppresses only in-process child subagents, drains only the shutting-down session's historian/recomp and the active owner's Dreamer jobs (cancels after ~5s), routes command output through the live RPC UI, and isolates the shared Dreamer lifecycle across plugin instances and worktrees.

**New Features**
- Scopes child-session detection to lifecycle `AsyncLocalStorage`, unregisters lifecycle listeners on `session_shutdown`, and keeps `agent_end` synchronous.
- Captures a session snapshot at command start, threads cancellation through recomp and memory migration, blocks late UI writes, and aborts long drains safely.
- Shows short progress via notifications and detailed results in dialogs, with a safe fallback to notifications; applied to `/ctx-status`, `/ctx-embed`, `/ctx-recomp`, `/ctx-session-upgrade`, `/ctx-wrapup`, `/ctx-flush`, and `/ctx-dream`.
- Claims process-wide startup tasks once and defers session-history reads until the backfill lease is acquired.
- Shares Dreamer registration per project with an active owner requirement for manual runs, transfers ownership on exit, rejects stale/ownerless requests, prevents stale runs across worktrees, and stops shared timers when idle.
- Tracks and cancels in-flight recomp/upgrade runs per session; waits only the matching session during shutdown and aborts remaining work after a short grace period.
- Resolves the subagent CLI safely in embedded hosts by reusing `process.argv[1]` only when the host is the Pi CLI, preferring packaged/bundled binaries otherwise, and never spawning with a shell.

<sup>Written for commit fd89658083ab316a8095dc3028d8e97e92927435. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

The PR adapts the Pi integration for multiple sessions sharing one RPC host process.
- Isolates in-process child initialization while allowing independent sessions to initialize.
- Coordinates process-wide startup maintenance and session-scoped background-job shutdown.
- Adds owner-aware Dreamer registration, manual execution, and worktree handoff.
- Routes command status through live RPC notifications and dialogs.
- Safely resolves the Pi CLI when running inside embedded hosts.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/pi-plugin/src/index.ts | Introduces child-context isolation, process-wide startup claims, per-session lifecycle cleanup, and owner-aware Dreamer wiring. |
| packages/pi-plugin/src/dreamer/index.ts | Adds process-shared project registrations with explicit owners, safe timer replacement, owner-bound manual runs, and scoped in-flight drainage. |
| packages/plugin/src/plugin/dream-timer.ts | Makes shared timer registration cleanup identity-aware so stale cleanup cannot remove a replacement registration. |
| packages/pi-plugin/src/pi-recomp-runner.ts | Tracks detached recomp and upgrade work per session with cancellation and guarded cleanup. |
| packages/pi-plugin/src/commands/pi-command-utils.ts | Preserves model-invisible command entries while presenting status through live RPC notifications and modal dialogs. |
| packages/pi-plugin/src/subagent-runner.ts | Hardens executable resolution so embedded hosts do not accidentally relaunch their own server entrypoint. |
| packages/pi-plugin/src/commands/ctx-recomp.ts | Captures session state before detached execution, propagates cancellation, and defers marker and context refresh effects safely. |
| packages/pi-plugin/src/commands/ctx-session-upgrade.ts | Applies the same snapshot and cancellation lifecycle to recompilation and memory migration during session upgrades. |

</details>

```mermaid
flowchart TD
  P[Persistent Pi RPC host] --> S1[Session A extension]
  P --> S2[Session B extension]
  P --> C[In-process child session]
  C --> G[AsyncLocalStorage child guard]
  G -->|Child context| N[Suppress full initialization]
  S1 --> M[Shared process maintenance]
  S2 --> M
  M --> O[Run startup work once]
  S1 --> D[Shared Dreamer project registration]
  S2 --> D
  D --> T[Active owner timer]
  S1 --> J1[Session-scoped detached jobs]
  S2 --> J2[Session-scoped detached jobs]
  J1 --> U1[Session A RPC UI]
  J2 --> U2[Session B RPC UI]
```
</details>

<sub>Reviews (10): Last reviewed commit: ["Merge branch &#39;master&#39; into fix/pi-web-co..."](https://github.com/cortexkit/magic-context/commit/fd89658083ab316a8095dc3028d8e97e92927435) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55432287)</sub>

<details><summary><h4>Context used (5)</h4></summary>

- Knowledge Base — [Pi plugin runtime](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/pi-plugin.md)
- Knowledge Base — [Runtime integrations](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/runtime-integrations.md)
- Knowledge Base — [Historian and dreamer workflows](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/historian-and-dreamer.md)
- Knowledge Base — [Dreamer orchestration](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/dreamer-orchestration.md)
- Knowledge Base — [Historian production flow](https://app.greptile.com/cortexkit/-/custom-context/knowledge-base/cortexkit/magic-context/-/docs/historian-production.md)
</details>


alfonso-magic-context (changes requested):
Thanks for this — especially as a first contribution. The diagnosis is right: the #247 process-global latch is what makes the second pi-web session skip Magic Context entirely, and routing child suppression through `subagents:child:session-created`/`disposed` plus AsyncLocalStorage is the correct seam. Once-per-process startup maintenance, not reusing a non-Pi `argv[1]`, Dreamer sibling ownership, and keeping `ctx-status` entries model-invisible while presenting them in RPC are all the right instincts. And noted that you already pushed the Dreamer owner-handoff stabilization mid-review — that resolves one of the items we had flagged, and that kind of responsiveness makes this easy to shepherd.

Two clarifications so we don't talk past each other:
1. "RPC hosts" here is Pi RPC *mode* (`ctx.ui.notify`/`ctx.ui.custom`). It does not change Magic Context's RPC server, which must stay on `127.0.0.1` with a bearer token. We checked; this PR does not touch that.
2. The old "second init in the process is a no-op" test *should* change — that contract is the bug for pi-web. Please keep the child-only skip test (you did).

Before we can merge:
- Dreamer `registeredProjects` on `globalThis` (same jiti `moduleCache:false` reason as the child marker), so two sessions in one repo don't start two timers.
- `session_shutdown` draining only that session's in-flight work — in pi-web, shutdown is not process exit.
- RPC presentation using the command's live `ctx`, not a `session_start` closure.
- The #177 "never spawn bare `pi`" test kept *alongside* the new embedded-host test.
- `packages/pi-plugin/PARITY.md` updated for RPC dialogs, the multi-session process model, and the latch → ALS change.

We've approved CI for this PR so your next push gets the full check suite. Really solid work — happy to re-review quickly.

elrond298:
Rebased onto the latest master (`52e650c6`) and force-pushed the updated branch. No functional changes were made as part of the rebase.

alfonso-magic-context (changes requested):
Thanks for the rework — the quality jump since the last round is real, and most of it verified clean under a full re-review (all five prior items confirmed addressed at source; ALS suppression held up under a four-concurrent-child probe with an async hop; both suites green on the PR merged onto current master; no public adapter surface rides in, which matters because the API question is deliberately deferred to #353).

Four items block the merge, two of them reproduced by probe rather than read off the diff:

1. **Stale sibling-worktree dreamer timer.** `registerPiDreamerProject` keeps the old owner in the shared owners map when a different owner re-registers the same project identity from another directory (`src/dreamer/index.ts:157-185`), and the retired timer's predicate checks only that old owner's entry and directory (`:190-194`) — not the current registration generation/activeOwner. A probe with owner A/worktree A then owner B/worktree B showed the old A client still prompts. The committed test (`dreamer/index.test.ts:428-470`) covers one owner changing directories and misses this. Bind scheduled clients to the current registration generation (or activeOwner+projectDir) and add the sibling-owner regression.

2. **Shutdown timeout abandons rather than cancels.** `src/index.ts:2337-2373` stops waiting after 5s while `timeout.ts:1-14` leaves the underlying promise alive; detached recomp/upgrade retains `ctx` (`commands/ctx-recomp.ts:177-278`) and `pi-recomp-runner.ts:69-84` calls `onStatusChange` from `finally` without a stale-context guard. Pi 0.83 invalidates command contexts after disposal, so long recomp work can outlive shutdown and throw on dead ctx/UI. Capture immutable session data up front, fence or cancel on shutdown, and guard post-await presentation.

3. **Native Pi 0.83 RPC dialogs are silently vacuous.** `pi-command-utils.ts:180-190` assumes `ctx.ui.custom` renders or rejects, but Pi 0.83 rpc-mode implements it as `Promise.resolve(undefined)` — the catch fallback never runs, so native/external RPC hosts lose the detailed results entirely. It works in your host because pi-web supplies its own custom UI. Add a capability check with a portable notification fallback (and a test against the real 0.83 shape, not a functional fake), or scope the PARITY.md claim explicitly to hosts that provide `ui.custom`.

4. **Missing #247 storm regression.** The process-global latch you removed originally existed for the four-child in-process init storm. Our probe of your ALS scoping passed it — the implementation looks right — but no committed test reproduces the storm. Please add the parallel multi-child regression asserting every child registers no tools/events/background scans.

Nothing else stands between this and merge — the internals-first direction is settled on our side, and #353 tracks the public API question separately.

elrond298:
Thanks for the detailed review. I addressed the four remaining items, rebased the branch onto `master` at `7d6fda35`, and force-pushed the updated commits.

1. **Dreamer timers and results from an old registration**

   Each Dreamer registration now has a version number. A scheduled Dreamer job checks that version before creating a session, sending a prompt, reading messages, and accepting the final result. If another worktree or a reloaded plugin has replaced the registration, the old job stops and its result is ignored.

   I also covered the asynchronous A → B → A case. If the first A timer finishes starting after the final A timer has already been registered, cleanup from the first timer cannot remove the final timer entry stored for that project directory.

2. **Recompilation and upgrade tasks after session shutdown**

   Recompilation and session upgrade now copy the required session, directory, branch, and model data when the command starts. Their background tasks no longer read from the command context after it may have become invalid.

   During `session_shutdown`, the plugin waits up to five seconds for that session’s background tasks to finish. Tasks that are still running then receive a cancellation signal. After cancellation or shutdown, their results cannot update compaction markers, migration state, memories, or the session’s UI.

   `agent_end` still returns immediately. Waiting and cancellation happen only during `session_shutdown`.

3. **Dialog fallback in native Pi 0.83**

   Native Pi 0.83 can resolve `ui.custom()` with `undefined` without displaying anything or throwing an error. The plugin now records whether the rendering callback passed to `ui.custom()` was actually called.

   pi-web continues to display the custom dialog. If native Pi does not call the rendering callback, the command displays the same information as a notification instead.

   All commands that display status also receive the session shutdown signal. If `ui.custom()` resolves or rejects after shutdown, the command does not use the old UI context to display a notification.

4. **Four child sessions initializing concurrently**

   I added a behavioral test that starts four child sessions at the same time and closes them in different orders, including while other child sessions are still initializing.

   The test verifies that:

   - each child session keeps its own child-session state;
   - child sessions do not register primary-session tools, event handlers, entry renderers, or startup maintenance;
   - closing one child session does not change how the other child sessions are handled;
   - the parent session still performs its normal initialization.

   This keeps the current `AsyncLocalStorage` design and verifies that concurrently running child sessions do not overwrite one another’s state.

Additional points from the earlier review remain covered:

- All plugin instances in the same process use the same Dreamer registration map, preventing duplicate timers for the same project.
- Shutdown waits only for work owned by the session and Dreamer registration being closed.
- RPC commands receive the current command context when invoked instead of retaining a context from plugin startup. Their status output is displayed in the UI and is not added to the model conversation.
- The #177 regression remains a separate test. It creates a valid Pi CLI path and verifies that the child process uses `process.execPath` with that `cli.js`, does not execute bare `pi`, and does not use a shell. The adjacent embedded-host test separately verifies that the plugin does not re-run an unrelated host program from `process.argv[1]`.
- `PARITY.md` documents the shared process behavior, RPC presentation behavior, and how `AsyncLocalStorage` keeps concurrently initializing child sessions separate.
- No implementation of the Host Adapter API proposed in #353, and no dashboard changes, are included in this PR.

After rebasing onto `7d6fda35`, the Pi-plugin suite passes 848 tests. Another 59 tests covering the changed Dreamer and shared code also pass. The build, workspace type checking, formatting check, and Git whitespace check pass.

The new CI and Smoke runs are currently waiting for workflow approval. Cubic reports that it skipped automatic review because the force-push rewrote the branch history, so it will need to be triggered manually.

ualtinok:
Reviewed the lifecycle rework (head 1425715) — strong pass. Three of the four requested changes are cleanly addressed and well-tested: the RPC dialog's factory-invocation tracking with the notify fallback, the AsyncLocalStorage child marker replacing the process-global latch (independent sessions now initialize normally — nice resolution of the #247 tension), and the fenced per-session shutdown drain with owner-scoped awaits.

One item still open before this can merge, on blocker 1: the stale-timer fix fences the *client* by generation, but timer *cleanup* identity is still directory-only. In the A→B→A sequence, the first A registration's canceled timer sees the final registration also at directory A, skips `timerCleanup`, and the old timer resource stays registered even though its client is fenced — and the new test currently pins that skip as expected. Cleanup needs the same generation identity the client checks (drop the resource when `generation !== current`, not when the directory differs), and the A→B→A test should assert the first timer's cleanup *runs*.

Two smaller asks alongside: a shutdown integration test with two live sessions (retire A, assert B's historian/dreamer continue working), and a fresh CI run on the current head (the force-push left the checks neutral/empty). With those, this is mergeable — the rework's shape is exactly right, and the new owner/generation surfaces read well.

Issue #280, opened 2026-08-06
Migrations have no downstream lane — every release that adds one collides with fork migrations


iceteaSA:
### What problem does this solve?

`schema_migrations` is keyed on a single integer owned by upstream. Anything downstream that needs a schema change — a fork, or a sibling plugin sharing `context.db` — has to pick a number out of the same sequence, and upstream's next release takes it.

We track this repo closely and have renumbered six times: v33 → v41 → v51/52/53 → v70/71/72 → v73/74/75 → v76. Mechanical, but it is not free, because the collision is silent in the worst case.

The failure mode is the part worth fixing. A downstream migration records itself as, say, v41. Upstream then ships its own v41. On the next start the runner sees version 41 already applied and **skips upstream's body**. No error. The columns simply never exist, and you find out when a query throws much later:

```
classify chunk failed: SQLiteError "no such column: shareable"
```

That specific one ran for days before surfacing. We have hit this class three separate times, each needing manual DB surgery to repair (delete the colliding rows, re-run the real bodies). The `ensureColumn` self-heal net in `initializeDatabase` catches some of it, but only for tables that have a net — migration-only tables have no backstop at all.

To be clear about scope: this is a downstream problem, not an upstream bug. Nothing here is broken for people running stock. Filing it because the cost lands on anyone extending the plugin, and the fix is cheap if you want it.

### Proposed solution

Two options, in the order I would pick them:

**Reserve a downstream range.** Upstream keeps `< 10000`; forks and sibling plugins use `>= 10000`. One documented convention, no schema change, no runner change. Solves collision entirely as long as both sides honor it.

**Or key the table on `(owner, version)`.** `schema_migrations` gains an `owner TEXT NOT NULL DEFAULT 'upstream'`, and the runner tracks a high-water mark per owner. Structural, so it cannot be violated by convention drift, at the cost of a migration to the migrations table.

Either way, the fence check in `LATEST_SUPPORTED_VERSION` would need to compare per-owner rather than against a single max.

### Alternatives considered

- **Status quo.** Workable, and we have absorbed it six times. The objection is the silent-skip case, not the renaming.
- **Make the runner refuse to skip.** Verify a recorded migration's effect before skipping. Robust, but expensive on every open, and it needs a per-migration assertion.

### Additional context

Happy to prototype the range convention as a docs-only PR if that is the direction you like — it is the lower-risk of the two and needs no code.

alfonso-magic-context:
[...]

PR #382, opened 2026-08-28
fix(test): resolve the store version from the upstream lane in persistedVersion

iceteaSA:
## What

`persistedVersion` in `storage-db-initialize-order.test.ts` reads the store's schema version with a bare `MAX(version)`. Production doesn't — `getPersistedSchemaVersion` bounds the same read by the reserved fork lane:

```ts
// storage-db.ts:269-274
const row = db
    .prepare("SELECT COALESCE(MAX(version), 0) AS version FROM schema_migrations WHERE version < ?")
    .get(FORK_MIGRATION_VERSION_FLOOR) as { version: number } | undefined;
```

So for a downstream fork recording its migrations in the documented `>= 10_000` lane, production resolves the version correctly while this test helper reads the fork's row as the store version:

```
opens the repaired legacy store through the shared OpenCode and Pi boot paths
  Expected: 82
  Received: 10100
```

The test is correct in what it asserts — it's the helper that doesn't match the production rule it's standing in for.

## Change

Mirror the production bound in the helper, plus a regression that seeds a fork-lane row and asserts the resolved version is still `LATEST_MIGRATION_VERSION`.

Red-check: reverting the helper bound while keeping the new test gives `Expected: 82 / Received: 10100` — the same symptom, now caught directly rather than as a side effect of an unrelated boot-path assertion.

```
with the fix              5 pass / 0 fail
helper reverted, test kept 4 pass / 1 fail
```

## Scope

One test file, +35/−2. No production code touched.

`tsc --noEmit` clean. Lint reports the same 6 pre-existing errors as clean `master` (verified by stashing), none in this file.

## Why it matters

This is the only place I found where the reserved-lane convention from #280 isn't honoured. Production code, the schema fence, and the migration runner all respect the floor; this helper was the one straggler. Without it, any fork using the lane as documented sees a red suite on an upstream test that has nothing to do with its changes — which is exactly the friction the reserved lane was introduced to remove.

Found while rebasing a fork onto `master` after #376/#378 merged: the failure looked like one of the known full-suite concurrency artifacts, but it reproduced in isolation (2.6s, single file) and passed on clean `master`, which is what pointed at the helper rather than the environment.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/382"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790546500&installation_model_id=22398&pr_number=382&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F382&signature=75fb5045bb5a179165db048b714e64e27c941e93752865204132521672f3ddbe"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes the `persistedVersion` test helper so it resolves the store version from the upstream migration lane, matching production's `getPersistedSchemaVersion`. Previously the helper used a bare `MAX(version)` query, so a downstream fork recording migrations in the reserved `10_000+` lane was treated as the store version and failed the shared boot assertion.

- Adds a regression test that seeds a fork-lane row and asserts the resolved version is still `LATEST_MIGRATION_VERSION`.

<sup>Written for commit 2580291fe0455c997394bc38a2e449af12f8b604. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/382?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

This PR aligns the test-only `persistedVersion` helper with production by excluding migration rows in the reserved downstream-fork lane.

- Imports and applies `FORK_MIGRATION_VERSION_FLOOR` to the helper query.
- Adds regression coverage proving a fork-lane row does not change the resolved upstream schema version.
</details>

The PR appears safe to merge, with no actionable defects identified in the test-only change.

The updated query uses the same constant and strict boundary as production, while the regression fixture remains isolated and is cleaned up through the existing database lifecycle.

| Filename | Overview |
|----------|----------|
| packages/plugin/src/features/magic-context/storage-db-initialize-order.test.ts | The helper now mirrors production’s bounded schema-version query, and the added isolated regression test correctly covers the downstream fork-lane case. |

</details>

<sub>Reviews (2): Last reviewed commit: ["fix(test): resolve the store version fro..."](https://github.com/cortexkit/magic-context/commit/2580291fe0455c997394bc38a2e449af12f8b604) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58082791)</sub>

### DIFF
--- a/packages/plugin/src/features/magic-context/storage-db-initialize-order.test.ts
+++ b/packages/plugin/src/features/magic-context/storage-db-initialize-order.test.ts
@@ -6,7 +6,12 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { Database } from "../../shared/sqlite";
 import { closeQuietly } from "../../shared/sqlite-helpers";
-import { LATEST_MIGRATION_VERSION, MIGRATIONS, runMigrations } from "./migrations";
+import {
+    FORK_MIGRATION_VERSION_FLOOR,
+    LATEST_MIGRATION_VERSION,
+    MIGRATIONS,
+    runMigrations,
+} from "./migrations";
 import { closeDatabase, initializeDatabase, openDatabase, openDatabaseAsync } from "./storage-db";
 
 interface InitializerIndexAudit {
@@ -137,8 +142,16 @@ function seedHistoricalStoreFile(version: number): { dir: string; path: string }
 }
 
 function persistedVersion(db: Database): number {
+    // Mirror getPersistedSchemaVersion: rows at or above the reserved fork lane are a
+    // downstream namespace, not this binary's schema version. A bare MAX(version) reads
+    // a fork's 10_000+ row as the store version, so this assertion fails for any fork
+    // using the documented lane even though production resolves the version correctly.
     return (
-        db.prepare("SELECT COALESCE(MAX(version), 0) AS version FROM schema_migrations").get() as {
+        db
+            .prepare(
+                "SELECT COALESCE(MAX(version), 0) AS version FROM schema_migrations WHERE version < ?",
+            )
+            .get(FORK_MIGRATION_VERSION_FLOOR) as {
             version: number;
         }
     ).version;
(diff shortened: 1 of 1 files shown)


iceteaSA:
Pushed one correction, caught by running this against a real fork rather than only against clean `master`.

The regression test seeded its fork-lane row with a plain `INSERT` at `FORK_MIGRATION_VERSION_FLOOR + 100`. That passes on `master` (no fork rows) but fails on an actual fork whose own migration already occupies that version:

```
SQLiteError: UNIQUE constraint failed: schema_migrations.version
```

Which is a poor property for a test whose entire purpose is to protect forks using the lane. Changed to `INSERT OR REPLACE` so the fixture is idempotent regardless of what the store already contains.

Verified both directions now:

```
clean master + fix      5 pass / 0 fail
fork (rows at 10_000-10_002, 10_100) + fix   5 pass / 0 fail
helper bound reverted   4 pass / 1 fail — Expected: 82, Received: 10100
```

Issue #416, opened 2026-09-03
[pi-plugin] OMP-native subagent spawning: run historian/dreamer/sidekick as OMP task subagents instead of --print subprocesses


Qiiks:
## Summary

On OMP (oh-my-pi) hosts, Magic Context currently spawns every subagent — historian, dreamer, sidekick, recomp, wrapup, upgrade — as a fresh headless CLI process (`omp --print --mode json --no-session …`, via `PiSubagentRunner`). OMP has a native subagent system (the same one that powers its `task` tool and Agent Hub / subagent pane), and running MC's subagents through it would give OMP users first-class visibility and lifecycle management that the invisible `--print` process model can't.

## What OMP's native subagent system provides

Verified against OMP main (`3e181fe0`):

- **Programmatic spawn API** — `runStructuredSubagent()` (`packages/coding-agent/src/task/structured-subagent.ts:556`) accepts `{session: ToolSession, invocationKind, assignment, agent, model, outputSchema, identity, enableIrc, enableLsp, signal, onProgress, keepAlive, detached}`. OMP itself uses this for non-model-initiated spawns (its Cleanse feature builds a synthetic `ToolSession` and calls it directly — `src/cleanse/agent.ts:95-180`). So no model `task`-tool call is required.
- **TUI subagent pane visibility** — spawned agents publish `task:subagent:lifecycle/progress` frames on the root-scoped `subagentEventBus` (inherited by every spawned session, `tools/index.ts:233-239`). The TUI's `SessionObserverRegistry` renders live progress rows from those frames (`modes/session-observer-registry.ts:145-246`, `renderSubagentHudLines` at `modes/interactive-mode.ts:507` — shows `kind === "subagent" && status === "active" && detached === true`). Users can watch the historian work in real time, open its transcript (`history://<id>`), and see it in Agent Hub.
- **Model-isolation guarantees** — the spawned child gets its tool set gated by `enableIrc: false` (no `hub` tool at all: no messaging, no roster, no cancel path from the child). Parent-side cancel is owner-scoped (`tools/hub/jobs.ts:396-398` — `ref.parentId !== ownerId` → "was not spawned by you and cannot be cancelled"), so the main agent cannot cancel or message an MC-spawned subagent it didn't spawn via its own `task` tool. The main agent also never sees the spawn announcement (those are delivered only through the parent tool-call path), so weaker main models can't identify or interfere with the historian.
[...]

Issue #21, opened 2026-04-17
Crash: SIGBUS/SIGTRAP in onnxruntime-node ReleaseIoBinding during local embedding load (concurrent opencode sessions, macOS arm64)


vinayakkulkarni:
## Description

Both opencode TUI sessions crash with a native SIGBUS / SIGTRAP inside `onnxruntime-node` while the magic-context plugin loads the local embedding model (`Xenova/all-MiniLM-L6-v2`). The crashes occurred in two concurrently-running opencode processes (separate sessions, separate cwd) within seconds of each other — strongly suggests a concurrency / double-free bug in the local embedding loader when two processes touch the same cached model at the same time.

## Environment

- Plugin: `@cortexkit/opencode-magic-context@0.9.1`
- OpenCode: `1.4.7`
- OS: `macOS 26.4.1` (build 25E253)
- Arch: `arm64` (Apple Silicon, M2 Max — `Mac14,6`)
- Node: `v24.15.0`
- `onnxruntime-node`: `1.21.0` (npm) — **bundles `libonnxruntime.1.14.0.dylib`** (native C++ lib)
- Transitive dep via: `@huggingface/transformers@~3.7.6`

## Configuration

`~/.config/opencode/magic-context.jsonc`:

```jsonc
{
  "$schema": "https://raw.githubusercontent.com/cortexkit/opencode-magic-context/master/assets/magic-context.schema.json",
  "enabled": true,
  "historian": {
    "model": "anthropic/claude-haiku-4-5",
    "fallback_models": ["opencode-go/glm-5"]
  },
  "dreamer": {
    "enabled": true,
    "model": "anthropic/claude-sonnet-4-6"
  },
  "sidekick": {
    "model": "anthropic/claude-haiku-4-5"
  },
  "experimental": {
    "user_memories": { "enabled": false },
    "pin_key_files": { "enabled": false }
  },
  "compaction_markers": false
}
```

No explicit `embedding` block, so the default (`provider: "local"`, model `Xenova/all-MiniLM-L6-v2`) is used.

## Reproduction

1. macOS 26.4.1 arm64, opencode 1.4.7, magic-context 0.9.1 (default local embedding config).
2. Cached model already present at `~/.local/share/opencode/storage/plugin/magic-context/models/Xenova/all-MiniLM-L6-v2/onnx/model.onnx` (90 MB, `md5 = 84f837de2a0f667784facf2ba0f36b22`).
[...]

PR #418, opened 2026-09-03
feat(pi-plugin): OMP-native subagent runner for the historian

Qiiks:
## Summary

Implements #416: on OMP (oh-my-pi) hosts, the historian now runs through OMP's **in-process structured-subagent machinery** (`runStructuredSubagent`) instead of a headless `--print` subprocess. Non-OMP Pi hosts keep the subprocess path unchanged — the swap is gated on `isOmpHostProcess()` plus a successful surface load.

## What this gives OMP users

Per the research in #416: the spawned historian is visible in OMP's subagent surfaces (Agent Hub / pane) when the host bus is reachable, is isolated from the main agent (`enableIrc: false` removes the child's hub tool entirely; an MC-initiated spawn is neither an owner job nor a `parentId: Main` registry child the model can cancel), and terminates via OMP's yield contract with the output mapped straight into MC's existing parsing.

## Design

**`omp-host.ts` — surface loader (memoized, never throws).** Resolves from the running OMP package, in priority order:

1. Walk up from `process.argv[1]`/`process.execPath` to the `@oh-my-pi/pi-coding-agent` package root, then import `src/task/structured-subagent.ts` directly (OMP ships its `src/` raw and maps subpaths via the `"./*"` exports wildcard, so this is the same module the host executes).
2. Walk up from this module's own install location (plugin dist sits under `~/.omp/plugins/node_modules/`, sibling of nothing by default — kept as a fallback for colocated installs).
3. Bare ESM subpath import of `@oh-my-pi/pi-coding-agent/task/structured-subagent`.

The surface carries three pieces: `runStructuredSubagent`, `Settings` (from `config/settings` — the synthetic ToolSession's required field, whose `reloadFromDisk()` the spawn policy calls), and `discoverAuthStorage` (from `./sdk`). Any failure resolves to `{ surface: null, reason }`; the wiring then constructs the existing `PiSubagentRunner`.

**`omp-subagent-runner.ts` — the runner.** Implements the harness-agnostic `SubagentRunner` contract from `@magic-context/core/shared/subagent-runner`:

- Builds the minimal synthetic `ToolSession` (required fields `cwd`, `hasUI`, `getSessionFile`, `getSessionSpawns`, `settings`, plus the isolation knobs: `enableIrc: false`, `enableLsp: false`, `enableMCP: false`, `restrictToolNames: true`, `taskDepth: 0`, `keepAlive: false`, `detached: true` — the same shape OMP's own Cleanse feature uses for non-model-initiated spawns).
- The historian's system prompt is threaded as a `<system_role>` block prepended to the assignment (OMP resolves system prompts from agent *definitions*; shipping `agents/*.md` is a fast-follow), with an explicit yield-contract instruction appended.
- Model refs translate through the existing `resolveModelRefForOmp` (same path the subprocess runner uses).
- Timeout races an internal `AbortController` against the caller's signal, with the pre-aborted-signal race handled (a signal that aborted before listener registration is honored immediately).
- Fail-soft result mapping per the contract: `timeout`/`abort`/`model_failed` (transient)/`non_zero_exit`/`no_assistant`/`spawn_failed` — never throws for run-level failures.
- Accounting parity: best-effort `recordChildInvocation` rows under a new `"omp"` harness value (`SubagentInvocationHarness` alias added; the CLI adapter types already had `"omp"`).

**Wiring (`index.ts`).** One shared `OmpSubagentRunner` singleton, constructed only when `isOmpHostProcess()`; `resolveHistorianFromConfig` swaps it in for the historian only. Recomp/wrapup/upgrade keep the subprocess runner this PR; converting them is mechanical follow-up once the seam is blessed. No config-key/schema changes in this prototype (opt-out via a `historian.spawn_mode` field comes with the general rollout).

## Verification

- **Unit**: 9 co-located tests covering surface-unavailable fallback (never throws), ok/failure/timeout/abort result mapping, transient hints, request-shape isolation flags, and the caller-abort race. Full pi-plugin suite failure set is byte-identical to pristine `master` (23 pre-existing Windows env failures).
- **Typecheck**: `tsc --noEmit` clean in both touched packages.
- **Live (OMP 18.1.5, Windows)**: a harness probe importing the installed dist resolved the surface from the real OMP package, constructed real `Settings`/`authStorage`, and executed a genuine structured spawn end-to-end — the child ran on `gmi-cloud/MiniMaxAI/MiniMax-M3`, yielded, and the runner mapped `{ ok: true, assistantText: "<test-output>…</test-output>", meta: { resolvedModel: "gmi-cloud/MiniMaxAI/MiniMax-M3:low", tokens: 3545 } }`.

**Known limitation (pre-decided in #416 discussion):** the extension context cannot reach the session's root `subagentEventBus`, so pane/HUD visibility of the spawn depends on the host-side surface proposed for OMP (`ctx.spawnSubagent`); without it the run is functionally complete (spawn, isolation, yield, output, accounting) but not rendered in the pane. The durable fix lives on the OMP side; this PR's value is the runner seam + native spawn semantics regardless.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/418"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791028762&installation_model_id=22398&pr_number=418&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F418&signature=0f0062242a1945c392233056db9214eda1c501eb19da6c2842c0b6d957d60969"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Runs the historian through OMP's in-process `runStructuredSubagent` on OMP hosts instead of a headless `--print` subprocess, implementing #416. Non-OMP Pi hosts keep the existing subprocess path, and runs fall back to it when the OMP surface can't load.

**Details**

- Isolates native runs from IRC, LSP, MCP, and persistent child sessions, then ends them via OMP's yield contract.
- Maps `thinking_level` to OMP's `lo`/`med`/`hi` effort selectors and exact `:level` model suffixes, including `off` with effort omitted, while guarding colon-qualified refs; `off` on a non-level-suffixed ref (e.g. `:free`) delegates to the subprocess runner.
- Threads the parent session's live ModelRegistry into structured spawns so extension-registered runtime providers resolve, and ensures the explicitly configured historian model wins over the host's ambient session model.
- Snapshots host context at run start, and refreshes it at session start and immediately before each spawn, so a session or model change mid-run never leaks into the in-flight spawn.
- Records every attempt under the `"omp"` harness after final classification, including empty-output and rejected preflight/deadline/abort failures.
- Recomp, wrapup, and upgrade subagents stay on the subprocess runner.
- Native output uses existing parsing but isn't rendered in the subagent pane yet.
- Tests cover fallback delegation, isolation, cancellation, result mapping, thinking-level translation, model precedence, registry forwarding, and mid-run context snapshots.

<sup>Written for commit 81fc79f38a691263dff50cf0ea85853bfec66bea. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/418?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

The PR adds an OMP-native historian runner while retaining the Pi subprocess runner as a fallback.
- Loads OMP’s structured-subagent surface lazily and safely.
- Captures live host model, registry, and session context at historian spawn time.
- Maps native results, cancellation, deadlines, thinking levels, and accounting into the existing runner contract.
- Corrects the previously reported fallback and accounting paths.
</details>

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| packages/pi-plugin/src/omp-subagent-runner.ts | Implements the OMP-native runner, subprocess fallback, context snapshots, model/thinking translation, final result classification, cancellation, and accounting. |
| packages/pi-plugin/src/omp-host.ts | Adds a memoized, fail-soft loader for the running OMP host’s structured-subagent surface. |
| packages/pi-plugin/src/index.ts | Selects the shared OMP historian runner on OMP hosts and captures live session context during startup. |
| packages/pi-plugin/src/context-handler.ts | Refreshes the OMP host snapshot immediately before each historian spawn. |
| packages/pi-plugin/src/omp-subagent-runner.test.ts | Covers native result mapping, fallback delegation, isolation flags, cancellation, model precedence, context snapshots, and thinking-level translation. |
| packages/plugin/src/features/magic-context/storage-subagent-invocations.ts | Extends invocation harness attribution to support OMP. |
| packages/plugin/src/features/magic-context/subagent-token-capture.ts | Accepts the shared OMP-capable invocation harness type for child accounting. |

</details>

```mermaid
flowchart TD
    A[Historian spawn requested] --> B{OMP host?}
    B -- No --> C[Pi subprocess runner]
    B -- Yes --> D[OMP runner]
    D --> E{Structured surface available?}
    E -- No --> C
    E -- Yes --> F[Capture host context]
    F --> G[Run structured subagent]
    G --> H[Classify final result]
    H --> I[Record invocation accounting]
    C --> J[Return shared runner result]
    I --> J
```
</details>

<sub>Reviews (9): Last reviewed commit: ["fix(pi-plugin): refresh omp host snapsho..."](https://github.com/cortexkit/magic-context/commit/81fc79f38a691263dff50cf0ea85853bfec66bea) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59934051)</sub>

### DIFF
--- a/packages/pi-plugin/src/context-handler.ts
+++ b/packages/pi-plugin/src/context-handler.ts
@@ -219,6 +219,7 @@ import {
 	resolvePiUsableContextLimit,
 	resolvePiWindowGeometry,
 } from "./pi-context-limit";
+import { captureOmpHostContext } from "./omp-subagent-runner";
 import { type PiHistorianDeps, runPiHistorian } from "./pi-historian-runner";
 import {
 	formatPiPressureForLog,
@@ -3645,6 +3646,11 @@ function spawnPiHistorianRun(args: {
 		}
 		const renewal = startPiCompartmentLeaseRenewal(db, sessionId, holderId);
 		try {
+			// Refresh the OMP runner's host snapshot from the live context right
+			// before spawning: session_start may not have fired for this session
+			// (resume paths), and /model switches mid-session would otherwise
+			// leave a stale registry/model. No-op on non-OMP runners.
+			captureOmpHostContext(historian.runner, ctx);
 			await runPiHistorian({
 				db,
 				sessionId,
--- a/packages/pi-plugin/src/index.ts
+++ b/packages/pi-plugin/src/index.ts
@@ -166,9 +166,11 @@ import { registerStatusLine, updateStatusLine } from "./status-line";
 import { stripTagPrefixFromAssistantMessage } from "./strip-tag-prefix";
 import {
 	configurePiSubagentExtensions,
+	isOmpHostProcess,
 	MAGIC_CONTEXT_PI_SUBAGENT_ENV,
 	PiSubagentRunner,
 } from "./subagent-runner";
+import { OmpSubagentRunner } from "./omp-subagent-runner";
 import {
 	buildMagicContextBlock,
 	clearPiSystemPromptSession,
@@ -675,6 +677,29 @@ export function resolveSidekickFromConfig(
 	};
 }
 
+
+/**
+ * Shared OMP-native historian runner, constructed at most once per process.
+ * `undefined` when this is not an OMP host — the caller then wires the
+ * subprocess runner. Construction is deliberately lazy and sync (the OMP
+ * surface load happens on first `run()`), so this never touches host modules
+ * at boot.
+ */
+let ompRunnerSingleton: OmpSubagentRunner | undefined;
+let ompRunnerResolved = false;
+
+function getOmpHistorianRunner(): OmpSubagentRunner | undefined {
+	if (!ompRunnerResolved) {
+		ompRunnerResolved = true;
+		if (isOmpHostProcess()) {
+			ompRunnerSingleton = new OmpSubagentRunner();
+		}
+	}
+	return ompRunnerSingleton;
+}
+
+const resolveOmpRunner = getOmpHistorianRunner();
+
 export function resolveHistorianFromConfig(
 	config: MagicContextConfig,
 ): PiHistorianOptions | undefined {
@@ -699,7 +724,14 @@ export function resolveHistorianFromConfig(
 	const fallbackModels = resolved.fallbacks;
 
 	return {
-		runner: new PiSubagentRunner(),
+		// On OMP hosts, run the historian through OMP's native subagent
+		// machinery (visible in the subagent pane, yield-terminated) instead
+		// of a headless `--print` process. Falls back to the subprocess
+		// runner whenever the OMP surface is unavailable (cortexkit/magic-context#416).
+		runner:
+			isOmpHostProcess() && resolveOmpRunner !== undefined
+				? resolveOmpRunner
+				: new PiSubagentRunner(),
 		model,
 		fallbackModels,
 		historianChunkTokens,
--- a/packages/pi-plugin/src/index.ts
+++ b/packages/pi-plugin/src/index.ts
@@ -170,7 +170,7 @@ import {
 	MAGIC_CONTEXT_PI_SUBAGENT_ENV,
 	PiSubagentRunner,
 } from "./subagent-runner";
-import { OmpSubagentRunner } from "./omp-subagent-runner";
+import { captureOmpHostContext, OmpSubagentRunner } from "./omp-subagent-runner";
 import {
 	buildMagicContextBlock,
 	clearPiSystemPromptSession,
@@ -1323,25 +1323,10 @@ async function startPiMagicContextRuntime(
 		const current = resolveCurrentProjectDeps(ctx);
 		syncCtxMemoryToolEnabled(pi, current.config.memory.enabled);
 
-		// Give the OMP-native runner the live host context: the parent session's
-		// ModelRegistry (so extension-registered runtime providers like bai/
-		// nous-portal resolve in structured spawns instead of falling back to a
-		// bundled-only fresh registry), plus the active model ref and session
-		// ids for fallback/lease wiring. Best-effort — a failure here must not
-		// break session start.
-		try {
-			resolveOmpRunner?.setHostContext({
-				model:
-					ctx.model !== undefined
-						? `${ctx.model.provider}/${ctx.model.id}`
-						: undefined,
-				modelRegistry: ctx.modelRegistry,
-				sessionFile: ctx.sessionManager?.getSessionFile?.() ?? null,
-				sessionId: ctx.sessionManager?.getSessionId?.() ?? null,
-			});
-		} catch (err) {
-			log(`session_start: omp runner host context capture skipped: ${String(err)}`);
-		}
+		// Give the OMP-native runner the live host context (parent registry so
+		// extension-registered runtime providers resolve in structured spawns,
+		// plus the active model ref and session ids). Best-effort.
+		captureOmpHostContext(resolveOmpRunner, ctx);
 
 		await handlePiCloneSessionStart(event, ctx, {
 			db,
--- a/packages/pi-plugin/src/index.ts
+++ b/packages/pi-plugin/src/index.ts
@@ -1323,6 +1323,26 @@ async function startPiMagicContextRuntime(
 		const current = resolveCurrentProjectDeps(ctx);
 		syncCtxMemoryToolEnabled(pi, current.config.memory.enabled);
 
+		// Give the OMP-native runner the live host context: the parent session's
+		// ModelRegistry (so extension-registered runtime providers like bai/
+		// nous-portal resolve in structured spawns instead of falling back to a
+		// bundled-only fresh registry), plus the active model ref and session
+		// ids for fallback/lease wiring. Best-effort — a failure here must not
+		// break session start.
+		try {
+			resolveOmpRunner?.setHostContext({
+				model:
+					ctx.model !== undefined
+						? `${ctx.model.provider}/${ctx.model.id}`
+						: undefined,
+				modelRegistry: ctx.modelRegistry,
+				sessionFile: ctx.sessionManager?.getSessionFile?.() ?? null,
+				sessionId: ctx.sessionManager?.getSessionId?.() ?? null,
+			});
+		} catch (err) {
+			log(`session_start: omp runner host context capture skipped: ${String(err)}`);
+		}
+
 		await handlePiCloneSessionStart(event, ctx, {
 			db,
 			signalPendingMarker: signalPiDeferredCompactionMarkerDrain,
--- a/packages/pi-plugin/src/omp-host.ts
+++ b/packages/pi-plugin/src/omp-host.ts
@@ -1,0 +1,248 @@
+/**
+ * OMP host surface loader.
+ *
+ * Locates the running OMP host's structured-subagent spawn API
+ * (`runStructuredSubagent`) so `OmpSubagentRunner` can spawn MC subagents
+ * in-process through OMP's native task machinery instead of headless
+ * `--print` subprocesses (see `omp-subagent-runner.ts` and
+ * cortexkit/magic-context#416).
+ *
+ * Dynamic imports are REQUIRED here, not style drift (same pattern as
+ * `dreamer/pi-session-api.ts`'s module ladder): the module specifier is
+ * genuinely runtime-selected — it is derived from the running OMP host's
+ * install location, and a static import of `@oh-my-pi/pi-coding-agent`
+ * would execute host-native addons at extension-bundle load time even on
+ * non-OMP hosts where the package may be absent. Runtime loading also lets
+ * a failure resolve to `null` (fallback to the subprocess runner) instead
+ * of failing the plugin boot.
+ *
+ * OMP ships its `src/` tree raw and maps package subpaths via the
+ * `"./*": { import: "./src/*.ts" }` exports wildcard, so
+ * `@oh-my-pi/pi-coding-agent/task/structured-subagent` resolves to the same
+ * source module the running CLI executes. The module must be imported
+ * in-process (inside the OMP extension host) — standalone imports fail on
+ * native-addon loading outside the host process, which is fine: this loader
+ * only runs inside OMP where the addon is already loaded.
+ *
+ * Resolution order (memoized, never throws):
+ *  1. Walk up from the running entry (`process.argv[1]`, then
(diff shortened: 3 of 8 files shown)


Qiiks:
Both P1s addressed in 3b45cb6d:

1. **Surface failure bypasses fallback** — the runner no longer dead-ends. When the surface cannot load on a detected OMP host, `OmpSubagentRunner.run()` now delegates to `PiSubagentRunner` (sticky per-process; announced via `onProgress` as `omp:fallback-subprocess` so traces distinguish the paths). Fixed inside the runner rather than the wiring gate so *every* caller gets the guarantee, not just the historian path.

2. **Empty output records success** — `recordAccounting` now runs only after final classification: exit-0-with-empty-output is recorded as `no_assistant` (failed), not `completed`.

Test updated to assert the delegation contract; 9/9 pass, tsc clean.

Qiiks:
Follow-up fix in 71518230: the thinking-level passthrough was wrong. OMP's task-spawn `effort` accepts only the positional selectors `lo`/`med`/`hi` (`TASK_EFFORTS`; `validateEffort` rejects literals, and `resolveTaskEffortLevel`'s switch silently no-ops on them) — passing `high`/`max` raw would have been rejected or silently dropped to default thinking.

Now `thinking_level` rides two OMP-sanctioned mechanisms:
1. `effort` — explicit literal→positional map (minimal/low→lo, medium→med, high/xhigh/max→hi; off/auto/undefined omitted)
2. an exact `:level` suffix on the model ref (`bai/glm-5.3-flash:high`) — OMP's model-pattern grammar resolves literal levels as `explicitThinkingLevel`

The positional selector resolves against the target model's own ladder (hi = whatever the model tops out at: high/xhigh/max), and the suffix carries exact per-model levels. 14/14 tests, tsc clean.

Qiiks:
Addressed the cubic round in 6599e48b:

1. **thinkingLevel `off` honored** — `off` is a real ThinkingLevel selector; it now rides the model ref as `:off`, which OMP resolves to `disableReasoning: true` on the child (parity with the subprocess path's `--thinking off`). `effort` stays omitted for off (the positional selector has no off rung).

2. **Hermetic fallback test** (id 3924536939) — added a `fallbackRunner` injection seam; the surface-unavailable test now injects a stub and asserts delegation via the `omp:fallback-subprocess` progress marker. No `pi` spawn, no hang risk, no host dependency.

3. **Colon-bearing refs guarded** — `nous-portal/meituan/longcat-2.0:free` + `max` no longer produces `:free:max`; a non-level suffix suppresses the literal append and thinking intent rides the positional effort selector instead.

14/14 tests, tsc clean.

Qiiks:
Correction commit f38c459e: the previous commit (6599e48b) described off→:off and the colon guard in its message but did not contain that code — only the hermetic fallback test seam landed. This commit ships the real changes:

- `off` in the suffix table → rides the ref as `:off` (→ `disableReasoning` on the child); `auto`/`inherit` stay excluded
- explicit three-case ref classification: clean ref → append `:level`; existing `thinking` suffix → replace; non-level suffix (`:free`) → id untouched, `effort` carries the intent (no `:free:max`)
- tests for all three cases + off

16/16 tests, tsc clean.

Qiiks:
Both findings from the 13:24/13:33 round addressed in 53eb14a1:

1. **greptile P1 (id 3925006192) — rejected runs skip accounting**: every catch-path return (preflight throw → spawn_failed, deadline → timeout, cancellation → abort) now records a subagent_invocations row, matching the classified-result recording the success path already does.

2. **cubic P2 (id 3925083820) — off + :free ref**: the structured path cannot express disable-reasoning for that ref shape (`:free:off` is invalid grammar; effort has no off rung). That combination now delegates to the subprocess runner (`--thinking off`). Tested: delegation asserted, native surface never invoked.

17/17 tests, tsc clean.

Qiiks:
Found and fixed the production failure behind 'subagent run failed (non_zero_exit): Error: No model selected.' (af53d3f5)

**Root cause**: `runStructuredSubagent` forwards `session.modelRegistry` into `runSubprocess` (structured-subagent.ts:433 → executor.ts `registryFromParent`), but the synthetic ToolSession built by OmpSubagentRunner never carried a registry. `runSubprocess` therefore constructed a fresh ModelRegistry that only knows bundled-catalog providers + models.yml discoverables. Extension-registered runtime providers (bai, nous-portal, kiosapi, apinex, tencent, tokenrouter) exist in-memory on the parent's registry only — a configured historian model from any of them (`bai/glm-5.3-flash:high` here) fails to resolve, the spawn silently falls back to the session default model, and throws "No model selected" when neither the pattern nor a default resolves. The fresh registry also can't read these providers' models.db rows: `#loadCachedDiscoverableModels` only iterates config-declared `#discoverableProviders`.

**Fix**: thread the parent's live registry through `OmpRunnerHostContext.modelRegistry` → `ToolSession.modelRegistry` (the sanctioned seam, tools/index.ts:333) → `runSubprocess` reuses it (`registryFromParent=true`). `session_start` captures `ctx.modelRegistry` + active model ref + session ids into the singleton runner via `setHostContext()`.

18/18 tests, tsc clean.

Qiiks:
Follow-up fix in 33d4781d plus verification of two review questions:

1. **Precedence bug caught before shipping**: capturing `ctx.model` into hostContext with the old `hostContext.model ?? options.model` ordering would have shadowed the configured historian model with the parent's ambient model on every run. Flipped to `options.model ?? hostContext.model` — explicit config wins, ambient is fallback-only. Test asserts `request.model` carries the configured `bai/glm-5.3-flash:high` while the session carries the parent registry.

2. **bai/\* availability on OMP**: confirmed against OMP's own cache via bun:sqlite — the `bai` row (46 models, authoritative, zero unrestorable headers) contains `glm-5.3-flash`, `qwen3.8-flash`, and `deepseek-v4-flash-vision-exp`. (`omp models` CLI renders nothing under non-TTY pipes in this environment — attempted plain, `--json`, provider-scoped, and PTY-hub forms, always exit 0 / 0 bytes — so the models.db probe plus the parent session's own fallback-chain validation knowing `bai/glm-5.3-flash` as a key is the evidence.)

3. **Fallback iteration**: MC-owned, not runner-owned — `runFallbackHistorianPass` retries through the same runner per candidate (`model: candidate.entry.model`, `thinkingLevel: candidate.entry.qualifier`), so the bai→bai-vision→opencode-zen chain all benefit from the parent-registry fix. The 13:59Z 'No model selected' was the last error after every candidate failed the same fresh-registry resolution.

18/18 tests, tsc clean.

Qiiks:
Two replies to the 15:15Z round:

**P1 (id 3925938807) — already addressed in the reviewed commit.** The claim that the runner 'prefers hostContext.model over options.model' does not match commit 33d4781d under review: omp-subagent-runner.ts:336 reads `const hostModel = options.model ?? host.model;` — options.model (the configured historian model) is primary, ctx.model is fallback-only, exactly the requested semantics. That ordering landed in 33d4781d itself ('configured model wins over ambient host model'), with a test asserting `request.model` carries the configured bai ref while the session carries the parent registry. No change needed; the finding describes the pre-flip ordering.

**P2 (id 3925938817) — valid, fixed in fc23201a.** `run()` now snapshots `this.hostContext` into a local at entry and uses the snapshot for model resolution and session construction, so a session_start firing during the settings/auth awaits can't mix the new session's registry/ids into the in-flight run. New test mutates the singleton mid-flight and asserts the request keeps the original registry + session id (the mutation strictly precedes the post-await reads, so the test discriminates: without the snapshot it fails).

19/19 tests, tsc clean.

Qiiks:
Robustness follow-up in 81fc79f3: the host snapshot is now captured at two points instead of one.

session_start alone is fragile — it may not fire for resumed sessions, and a mid-session /model switch would leave a stale registry/model ref (with the failure mode being the same silent fallback this PR fixes). The shared best-effort `captureOmpHostContext` helper (no-op on non-OMP runners) now runs both at session_start (index.ts) and immediately before `runPiHistorian` in `spawnPiHistorianRun` (context-handler.ts), so every structured spawn sees the live parent registry, model ref, and session ids.

53 tests across both suites pass, tsc clean.

Qiiks:
Fair verdict, and the reasoning is sound — thanks for the direct shape.

Summary of where this lands:

- **The engineering findings stand on this head**: loader doesn't reuse `pi-session-api.ts` guards (cubic #21–23/#28), 17.x compat gaps (`TASK_EFFORTS`/`:max` grammar), truncated→ok accounting, throwing onProgress, Biome errors, missing agents/*.md + spawn_mode opt-out. All real; no dispute.
- **The architectural point is the one that matters**: a private import of OMP internals keyed to whichever minor is installed is coupling you (rightly) won't carry. The `ctx.spawnSubagent` seam in #416 is the durable contract, and the historian-only-runner carve-out was heading toward a second permanent private import, not less.
- With that, closing rather than iterating a tenth round. The branch (`feat/omp-native-subagent-runner`) stays up with the nine-round fix history (catch-path accounting, hermetic seams, `off`/`:free`, model precedence, run-start snapshot) — useful reference material for whoever builds the adapter once OMP ships the surface.

Watching #416; the adapter PR comes when `ctx.spawnSubagent` lands. Thanks for the review time across all rounds.

Issue #419, opened 2026-09-03
Batch Pi transcript mutations to preserve provider prefix-cache stability


foreverwow001:
## Summary

On long, tool-heavy Pi sessions, independently scheduled transcript mutations can repeatedly change an old prompt prefix across adjacent provider calls. This causes avoidable prompt-cache misses even though each individual cleanup is correct.

The mutation sources observed on 0.41.2 include queued drops, heuristic/auto reclaim, reasoning cleanup, protected-tail/emergency publication, and M0/M1 history refresh. Fixing only delayed `ctx_reduce` operations is therefore insufficient.

## Evidence

From a long Pi session on 0.41.2:

- Calls with a fresh old-prefix mutation: 136, average cache-read ratio 46.6%; 121/136 below 70%.
- Calls with a stable old prefix: 138, average cache-read ratio 91.8%; only 10/138 below 70%.
- The session contained only two explicit `ctx_reduce` calls that day, so automatic mutation paths were material contributors.
- The longest near-zero-cache run was 18 adjacent calls; the longest below-70% run was 26 calls.

An isolated same-process Pi RPC reproduction used a 30-step tool chain plus 12 follow-up turns. A prototype common-exit batch gate reduced transform bust records from 39 to 3 and improved the final-20-call cache-read ratio from 96.1% to 98.2%, with all tool calls/results and instructions preserved.

## Suggested behavior

At the common Pi transform exit rather than in only one cleanup source:

1. Allow at most one fresh old-prefix mutation batch per real user turn at normal pressure.
2. Permit one additional batch when pressure enters a higher force band.
3. Keep the absolute 95% emergency path unrestricted.
4. Do not apply tiny pending batches merely because one protected tag aged out; wait for a material batch or higher pressure.
5. Continue deterministic replay of already-persisted drops/reasoning on every pass, but do not classify identical replay as a fresh bust.
6. Defer M0/M1 materialization within the same turn when a valid cached snapshot exists.

This preserves the existing fail-safe emergency behavior while giving the provider prefix a stable interval after each batch.

## Acceptance ideas

- One user turn with many tool calls and queued drops does not mutate the old prefix repeatedly.
- Protected-tail and emergency paths participate in the same batching decision.
- At >=95% pressure, cleanup remains unrestricted.
- No loss of current-turn tool results, unresolved errors, or user decisions.
- Session reopen/reload and deterministic replay remain byte-stable.

ualtinok:
[...]

PR #433, opened 2026-09-10
fix(pi): reduce OMP native inputs and optional Codex reasoning

SilentBless:
## What

Reduce OMP's native tool inputs and eligible old Codex reasoning without discarding the surrounding provider-native history.

- Update the selected `function_call.arguments` or `custom_tool_call.input` while preserving call identity, unrelated native-only reasoning and hosted results.
- Match uniquely identified native calls even when OMP synthesized a block ID because the captured native item has no ID.
- Reuse OMP's existing custom-input fallback instead of inventing a new custom-tool representation.
- Clear old encrypted reasoning only on Codex models whose resolved compatibility settings explicitly allow omission, and only from incremental history without protected native dependencies.
- Replay the same durable clearing decision and restore the original native payload if persisting its watermark fails.

Native text, user/developer carriers and full-history semantics are not rewritten. Ordinary tag injection does not mutate native history. Scheduling and protected windows are unchanged. This PR does not claim to solve every native-text reduction case.

## Why

OMP intentionally treats `assistant.providerPayload.items` as authoritative on same-model native replay. Magic Context currently reduces a tool input or clears old reasoning in `assistant.content` while leaving that native copy intact. The queue can drain and record a dropped tag, yet the original large input and encrypted reasoning are serialized into the next request.

Clearing the whole assistant payload is not an equivalent fix: a source-level counterexample removes the stale input but also loses unrelated native-only reasoning and a hosted image result, even though the selected tool pair remains valid.

The fix is deliberately limited to the two mutation units in the reproduction. Native-text synchronization was excluded after review identified fragmented text identities and instrumentation ownership concerns; it requires a separate contract, not a broader heuristic mapper here.

## Preservation boundaries

Native reasoning remains intact for non-Codex or unknown model capabilities, mandatory all-turn/tool-call reasoning, full snapshots, plaintext/malformed reasoning, redacted thinking and computer-linked reasoning. These cases are retained rather than speculatively reconstructed. Tool inputs with ambiguous identities are not guessed.

## Related work

- https://github.com/can1357/oh-my-pi/issues/592 and https://github.com/can1357/oh-my-pi/pull/594: related stale native replay after rehydration. Current OMP intentionally preserves Codex native history; this PR does not revive blanket payload invalidation.
- https://github.com/can1357/oh-my-pi/issues/4090: native-history durability is a compatibility constraint, not an issue this PR claims to close.
- https://github.com/cortexkit/magic-context/issues/419: cache-safe mutations must also cover the native representation.

No `Fixes` reference is used because these are related precedents, not demonstrated duplicates.

## Verification

The actual Magic Context queue/transcript functions and OMP serializers were exercised from their TypeScript sources using synthetic fixtures. The original queued reduction shrank its serialized request from 146,083 to 5,121 characters, removed old arguments and reasoning, and retained recent encrypted reasoning, a hosted image result and the selected tool pair. Additional source-level checks cover custom-input fallback parity, full-snapshot authority, synthesized tool IDs and mandatory reasoning preservation.

A subsequent live check used the normal installed OMP v18.1.10 CLI with an isolated profile and this built extension on `openai-codex/gpt-6-astra:xhigh`. Four model turns completed: the native tool input was actually replaced before a successful continuation, and a real encrypted reasoning item from a later response was then removed before another successful continuation (`MC_PROFILE_REASONING_OK.`). Its original ciphertext remained in the saved source session; the observer checked its hash was absent from outgoing native history, and the persisted clearing watermark advanced to its range. The native call ID remained stable. No `previous_response_id` shortcut or custom transport was used. This establishes live acceptance for this tested Codex flow, not every provider or native-history shape. Final gates after removing the standalone probes:

- Core: typecheck, lint, build, compiled-TUI check; **4,605 tests passed**.
- Pi: typecheck, lint, build; **1,036 tests passed**. Lint reports two existing warnings and no errors.
- Six source-serializer scenarios passed. The full OMP-mode context-handler fixture measured **54,932 → 4,947 characters**, with native text, the hosted result and the tool pair intact.
- A mixed plaintext/encrypted reasoning regression failed before its preservation guard and passes afterward. The rollback regression now asserts that the failing watermark write was actually attempted.

Independent reviews identified synthesized tool IDs, mandatory reasoning, mixed plaintext/encrypted items and native-text identity risks. The first three are handled by the targeted code and regressions; native-text rewriting was removed from scope. The separate `droppedTokens` accounting issue is unchanged.

<details>
<summary>Source-level OMP handler reproduction</summary>

Use sibling Magic Context and OMP source checkouts with their normal development dependencies (including the OMP native addon). The reference OMP revision was `34e938c3b37ab3b60375a6cef9c137166cb5fb70`. Save this inside `packages/pi-plugin/native-replay-probe.test.ts` and run `bun test native-replay-probe.test.ts` there. It runs the real context handler and serializer, not a remote provider.

```ts
import { expect, test } from "bun:test";
import { getPendingOps, getTagsBySession, queuePendingOp } from "@magic-context/core/features/magic-context/storage";
import { buildModel } from "../../../oh-my-pi/packages/catalog/src/build";
import { convertCodexResponsesMessages } from "../../../oh-my-pi/packages/ai/src/providers/openai-codex-responses";
import type { AssistantMessage, Context } from "../../../oh-my-pi/packages/ai/src/types";
import { createFakePi, createTestDb, fakeContext } from "./src/test-utils.test";
import { clearContextHandlerSession, registerPiContextHandler } from "./src/context-handler";
import { __setPiHarnessKindForTesting } from "./src/pi-harness-kind";

const model = buildModel({ id: "gpt-probe", name: "Probe", api: "openai-codex-responses", provider: "openai-codex", baseUrl: "https://example.invalid", reasoning: true, input: ["text"], cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 }, contextWindow: 272000, maxTokens: 64000 });
const zeroUsage = { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, totalTokens: 0, cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 } };
function assistant(content: AssistantMessage["content"], items: Record<string, unknown>[], dt = true): AssistantMessage {
  return { role: "assistant", api: model.api, provider: model.provider, model: model.id, content, providerPayload: { type: "openaiResponsesHistory", provider: model.provider, dt, items }, usage: zeroUsage, stopReason: "toolUse", timestamp: 2 };
}

test("full OMP context handler reduces the actual outgoing native history", async () => {
  const db = createTestDb();
  const sessionId = "full-omp-native-probe";
  try {
    __setPiHarnessKindForTesting("omp");
    const fake = createFakePi();
    registerPiContextHandler(fake.pi as never, { db, protectedTags: 20, heuristics: { clearReasoningAge: 20 } });
    const handler = fake.handlers.get("context") as (event: { messages: never[] }, ctx: never) => Promise<{ messages: never[] } | undefined>;
    const stale = "FULL_HANDLER_STALE_INPUT_".repeat(2000);
    const encrypted = "FULL_HANDLER_OLD_ENCRYPTED_".repeat(2000);
    const old = assistant([
      { type: "thinking", thinking: "old summary" },
      { type: "text", text: "Retain native text", textSignature: "msg_full" },
      { type: "toolCall", id: "call_full|fc_full", name: "read", arguments: { path: stale } },
    ], [
      { type: "reasoning", encrypted_content: encrypted, summary: [{ type: "summary_text", text: "old summary" }] },
      { type: "message", role: "assistant", id: "msg_full", content: [{ type: "output_text", text: "Retain native text", annotations: [] }] },
      { type: "function_call", id: "fc_full", call_id: "call_full", name: "read", arguments: JSON.stringify({ path: stale }) },
      { type: "image_generation_call", id: "ig_full", status: "completed", result: "aW1hZ2UtcGF5bG9hZA==", output_format: "png" },
    ]);
    const messages: Context["messages"] = [
      { role: "user", content: "Initial request", timestamp: 1 }, old,
      { role: "toolResult", toolCallId: "call_full|fc_full", toolName: "read", content: [{ type: "text", text: "old result" }], isError: false, timestamp: 3 },
    ];
    for (let i = 0; i < 25; i++) {
      messages.push({ ...assistant([{ type: "toolCall", id: `full_${i}|fc_${i}`, name: "read", arguments: { path: "example.ts" } }], []), providerPayload: undefined, timestamp: 4 + i * 2 });
      messages.push({ role: "toolResult", toolCallId: `full_${i}|fc_${i}`, toolName: "read", content: [{ type: "text", text: `result ${i}` }], isError: false, timestamp: 5 + i * 2 });
    }
    messages.push({ role: "user", content: "Continue", timestamp: 100 });
    const ids = messages.map((_, i) => `full-entry-${i}`);
    const run = async (percent: number) => {
      const input = structuredClone(messages);
      const result = await handler({ messages: input as never[] }, {
        ...fakeContext(sessionId, process.cwd(), ids, input as never), model,
        getContextUsage: () => ({ tokens: model.contextWindow * percent / 100, percent, contextWindow: model.contextWindow }),
      } as never);
      if (!result) throw new Error("context handler did not return messages");
      return result.messages as unknown as Context["messages"];
    };
    const initial = await run(0);
    const before = JSON.stringify(convertCodexResponsesMessages(model, { messages: initial }));
    expect(before).toContain(stale);
    const tag = getTagsBySession(db, sessionId).find(value => value.type === "tool" && value.messageId === "call_full|fc_full");
    if (!tag) throw new Error("full handler failed to tag old tool input");
    queuePendingOp(db, sessionId, tag.tagNumber, "drop");
    const reduced = await run(90);
    const items = convertCodexResponsesMessages(model, { messages: reduced });
    const after = JSON.stringify(items);
    expect(getPendingOps(db, sessionId)).toEqual([]);
    expect(after).not.toContain(stale);
    expect(after).not.toContain(encrypted);
    expect(after).toContain("Retain native text");
    expect(items.some(item => item.type === "image_generation_call")).toBe(true);
    expect(items.some(item => item.type === "function_call" && item.call_id === "call_full")).toBe(true);
    expect(items.some(item => item.type === "function_call_output" && item.call_id === "call_full")).toBe(true);
    console.log(JSON.stringify({ probe: "full-omp-context-handler", beforeWireChars: before.length, afterWireChars: after.length,
      nativeArgumentsAndOldReasoningRemoved: true, nativeTextHostedImageAndToolPairPreserved: true, providerRequestSent: false }));
  } finally {
    clearContextHandlerSession(sessionId);
    __setPiHarnessKindForTesting(undefined);
    db.close();
  }
});
```

</details>

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/433"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791611965&installation_model_id=22398&pr_number=433&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F433&signature=867301ce1f0e5b4939eff3d3dc7e5cad44dc81edb7337f91930583c833aa28f0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Fixes OMP native replay so tool-input reductions and eligible Codex reasoning clearing now mutate the serialized provider history and persist as durable replay state, so the reductions survive restarts and session clones.

**Bug Fixes**
- Rewrites the matched native `function_call.arguments` or `custom_tool_call.input` while preserving call identity, unrelated native-only reasoning, and hosted results; leaves calls untouched when reduced inputs cannot be safely serialized to JSON.
- Clears old encrypted reasoning only for Codex models whose resolved compatibility settings explicitly allow omission, and only from incremental history without protected native dependencies.
- Persists native tool inputs and reasoning ids in a versioned replay document that preserves trailing-blank and opaque namespace decisions; defers first activation to an authorized cache-busting pass and suspends saved decisions while compaction is off, resuming when re-enabled.
- Restores the original `providerPayload` if persisting replay state fails; malformed persisted state also fails closed, keeping native history intact, and the decision retries on the next authorized pass.
- Reuses OMP's existing custom-input fallback instead of introducing a new tool representation.
- Native text, user/developer carriers, and full-history snapshots are never rewritten; local Pi thinking cleanup still runs independently and advances only its own watermark, never the native reasoning set.
- Clone/session copy inherits only the retained native and trailing-blank decisions and remaps their ids to the clone.

<sup>Written for commit c27fc840ebc21c7e5beea2fc2643f6138f262aff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/433?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

This PR synchronizes Pi/OMP reductions with provider-native replay history and makes those reductions durable across restarts and cloned sessions.
- Rewrites selected native tool inputs while retaining call identity and unrelated provider-native items.
- Clears only eligible encrypted Codex reasoning under explicit compatibility and history-shape guards.
- Persists native replay decisions atomically and restores original payloads when persistence fails.
- Filters and remaps retained replay decisions when cloning session state.
</details>

The PR appears safe to merge; no actionable new defect remains after the clone replay remapping change.

Native mutations are compatibility-gated, persisted before publication, restored on write failure, and replayed deterministically. The latest clone change correctly filters decisions to retained assistant messages and remaps them through unique destination message IDs.

| Filename | Overview |
|----------|----------|
| packages/pi-plugin/src/native-replay-pi.ts | Safely rewrites uniquely matched native tool calls and conservatively clears eligible Codex reasoning. |
| packages/pi-plugin/src/native-replay-state-pi.ts | Persists and deterministically reapplies native tool-input and reasoning decisions. |
| packages/pi-plugin/src/context-handler.ts | Integrates native replay after transcript commit with compatibility, scheduling, and persistence-failure guards. |
| packages/plugin/src/features/magic-context/storage-replay-document.ts | Provides versioned parsing and serialization for combined trailing-blank and Pi-native replay state. |
| packages/plugin/src/features/magic-context/storage-clone.ts | Copies only retained replay state and remaps its message and tool identities into the clone. |
| packages/pi-plugin/src/clone-inheritance.test.ts | Covers retained-state filtering, identity remapping, legacy replay documents, and malformed-state rollback. |

</details>

```mermaid
flowchart TD
    A[Canonical Pi transcript] --> B[Apply local reduction]
    B --> C{Authorized cache-busting pass?}
    C -- No --> D[Replay previously persisted decisions]
    C -- Yes --> E[Derive native tool and reasoning decisions]
    E --> F{Persist replay state}
    F -- Success --> G[Rewrite provider-native payload]
    F -- Failure --> H[Restore original native payload]
    G --> I[Stable outgoing history]
    D --> I
    H --> I
    E --> J[Clone session]
    J --> K[Filter retained message IDs]
    K --> L[Remap replay decisions to clone IDs]
```
</details>

<sub>Reviews (6): Last reviewed commit: ["fix: remap trailing replay decisions whe..."](https://github.com/cortexkit/magic-context/commit/c27fc840ebc21c7e5beea2fc2643f6138f262aff) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=62469667)</sub>


SilentBless:
Review follow-up pushed in `b99a3fc3` (current head `10d797cc`, including a non-rewriting merge of upstream `18f6bb9a`).

- Guarded native function-call argument serialization. Circular data, BigInt, throwing `toJSON`, or a non-string serialization result leave the native payload unchanged instead of aborting the transform.
- Separated native-payload retention from local Pi thinking cleanup in both fresh clearing and durable replay. Ordinary stale thinking/signatures clear again; redacted blocks and protected native history remain intact.
- Gated native `providerPayload` rollback snapshots on `nativeReasoningMayClear`, removing unused snapshot/restore work on other models.
- Replaced the removed source-string watermark assertion with a behavioral inline-only persistence/restart test. It still replays the cleanup when the current age threshold prevents a fresh cleanup. A mutation check that ignored the inline watermark made the test fail by restoring the old `<think>` text.

The proposal to preserve every nonempty native `summary` was not applied. OMP's `finalizeReasoningThinking` maps `summary_text` to the ordinary Pi thinking display, which age-based cleanup intentionally removes. It is distinct from the protected native `content` payload. A regression and README clarification cover that distinction.

The README also states the native API boundary: OMP has no dedicated native-item mutation API. These reductions operate on the request transcript; the adapter does not itself perform a stored JSONL rewrite. OMP normally clones context input, with a shallow-array fallback for non-cloneable data; native tool-input updates are copy-on-write.

Verification on the updated upstream integration:
- Core typecheck, lint, build, compiled-TUI check: passed; **4,611 tests passed**.
- Pi typecheck, lint, build: passed; **1,040 tests passed**, with the existing lint warnings only.
- An additional request-copy/native subset passed **28 tests** after the documentation follow-up.

No new live model calls were made for this follow-up. The four-turn live evidence above remains tied to `77f8dda4`. The suggested reasoning-only and aborted-turn live cases have not been claimed as executed.

SilentBless:
Addressed the upgrade-state feedback in f16139294d46bc4c4e69cc70459ac7db6307e272.

- Legacy dropped tags and `clearedReasoningThroughTag` no longer authorize first native application during defer. Native tool-input values and native reasoning removals now have separate persisted state.
- Transcript setters collect input changes; the final native stage replays saved decisions, plans newly authorized changes, persists them, then publishes the new bytes. Failed writes retain the previous native representation. A later authorized input reduction can update the frozen value; defer cannot.
- Ordinary typed/inline thinking cleanup remains independent. Native-only reasoning uses its own replay state rather than advancing the local watermark. Positional `pi-msg-*` and synthetic IDs cannot own durable native reasoning decisions.
- Clone inheritance filters/remaps these records. The CLI still supports older schemas, and its later generic metadata copy cannot overwrite the filtered native decisions.
- README explicitly distinguishes function-call dropped-marker JSON from custom-call empty-string fallback; no identical copied-input rejection behavior is claimed.

Verification:
- Real OMP TypeScript serializer (34e938c3b37ab3b60375a6cef9c137166cb5fb70): five scenarios passed, including legacy upgrade, independent persistence failures and frozen-input progression. The upgrade sequence was A→A→A (653 chars, identical SHA-256), followed by the authorized transition and B→B→B (514 chars, identical SHA-256).
- Additional regressions cover native-only reasoning, stable-ID adoption, failed replacement of an already frozen input, restart replay, and clone metadata ownership.
- Core: 4,618 tests passed; typecheck, lint, build and compiled-TUI check passed.
- Pi: 1,047 tests passed; typecheck, lint and build passed. Only the two pre-existing lint warnings remain.
- Real Node SQLite smoke passed; the new native storage commit/rollback path was also exercised under Node.
- Two independent source reviews have no remaining blockers after the identity and clone fixes.

No new live-provider requests were made for this update. The earlier live Codex continuation evidence remains separate from these upgrade/cache-state checks.

SilentBless:
Implemented the preferred no-schema-migration shape in `671b2ac5` and `c27fc840`, including a merge of upstream `4bf838ff` that resolves the import conflicts without rewriting PR history.

- Removed the two `pi_native_*` initializer columns and their production consumers. The schema fence remains v84.
- `trailing_blank_decisions` now supports its existing flat v1 map and a v2 document containing `trailingBlank` and `piNative: { toolInputs, reasoningIds }`. Reads and no-op writes do not upgrade a flat document; native activation introduces v2.
- Blank add/demote and both native writers share whole-document CAS. Retries preserve the other namespaces. Tool inputs retain their exact serialized JSON; reasoning IDs remain a union. Unknown document versions fail closed; blank reads preserve valid legacy decisions, and malformed optional native state does not abort ordinary local cleanup.
- Clone inheritance owns the document so the later generic metadata copy cannot overwrite its filtered/remapped state. Local Cubic caught an additional mistake in the first implementation: the whole document was copied but trailing-blank IDs were left in the source identity space. The final clone path filters/remaps both known per-message decision sections, while preserving opaque sibling namespaces. Legacy-v1 and v2 clone regressions failed before that fix and pass afterward.

Verification on the final head:
- Core: **4,667 tests passed**, typecheck, lint, build, compiled-TUI check, and the real node:sqlite backend smoke passed.
- Pi: **1,073 tests passed**, typecheck, lint, and build passed; only the two pre-existing lint warnings remain.
- Additional real node:sqlite smoke exercised v84/flat upgrade, exact native input preservation, two-connection CAS interference, namespace preservation, and failed-write retention.
- Local Cubic re-review of the complete branch returned `issues: []`.

The incoming upstream tree had import-order and redundant-label lint failures in three files; only Biome's safe fixes were applied there, with no quality configuration changes. README/translations remain outside the PR change.

Compatibility boundary: updated readers accept both formats. This does not claim that a pre-patch binary can concurrently write the same session document; processes using that session should be restarted/updated together. No SQL migration, schema-fence bump, or new live-provider requests were needed.

PR #450, opened 2026-09-14
Pi: task-requested memory injection via ⟦mc-mem: …⟧ markers (v1)

### LINKED ISSUE #449: Feature Request: task-requested memory injection via ⟦mc-mem: …⟧ markers (Pi v1)
# Feature Request: task-requested memory injection via `⟦mc-mem: …⟧` markers (Pi v1)

## Motivation

When an orchestrator parent session already knows exactly which memory IDs a
subagent task needs, the current options are all indirect: the child can
auto-search (embedding cost, fuzzy ranking, no guarantee the right memory
surfaces), call `ctx_search` / `ctx_memory get` itself (requires the child to
know it needs to, and burns turn latency), or paste the full text into the
task (duplicates bytes, drifts from the memory store).

What is wanted is **deterministic, sitewise, budget-external delivery**: name
the IDs in the task, get the full memory content next to the task, before the
first token is generated. No embedding call, no retrieval ranking, no tool
round-trip, and unaffected by the `injection_budget_tokens` budget that may
have excluded exactly the memories the parent wants back.

This is the pi-subagents orchestration case: `defaultContext=fresh` children
share the project memory store but arrive with no parent conversation, so a
stable ID-addressed channel matters.

## Proposed behavior

A **real user task message** carrying a marker like:

```
Please follow the deploy SOP ⟦mc-mem: 11, 55⟧
```

gets the full content of memories 11 and 55 appended to the **end of that
message**:

```xml
<ctx-subagent-inject>
Task-requested memory snapshot (requested IDs: 11, 55; first read at 2026-09-14T10:00:00.000Z).
- [ID 11] [PROJECT_RULES] deploys happen only via the staging gate ...
- [ID 55] [CONFIG_VALUES] dashboard token budget is 8000 ...
</ctx-subagent-inject>
```

IDs that cannot be provided (missing, archived, expired, out-of-scope) render
one opaque `Unavailable` note per ID group — no content, category, or project
metadata leaks for memories the caller cannot see.

## Design summary

- **Marker safety**: bounded single-line locator (fixed character class, no
  nested quantifiers — linear time even on adversarial input; the naive
  regex measured >1500 ms at 26 digits, this pattern is O(n)) plus strict
  per-token validation: positive safe integers, optional leading `#`,
  separated by spaces/tabs/ASCII commas. Malformed markers are abandoned
  wholesale, never partially parsed. Full-width punctuation and look-alike
  brackets are never normalized.
- **Recognition scope**: only real user task text. Fenced code, inline code,
  blockquotes, already-generated augmentation blocks, assistant/tool/
  synthetic messages are ignored; markers split across text parts are never
  stitched; memory bodies inside our own payload are never re-parsed. To
  show the marker literally, wrap it in backticks.
- **Limits**: message-level merge, first-seen order dedupe, constant 10-ID
  per-task cap; over-limit IDs are reported as unprocessed, never silently
  dropped or back-filled.
- **Two-layer read filter**: access scope reuses the existing visibility
  rules (project/workspace identity, alias expansion, share categories) via
  a narrowly extracted shared helper; injection eligibility additionally
  requires `active`/`permanent` status and no expiry.
- **Snapshot semantics**: first read freezes availability and the read
  timestamp; the rendered payload is persisted and replayed byte-for-byte
  for as long as the original message stays visible. Later
  update/archive/expire does not refresh or retract history; a new task
  naming the same ID re-reads at its own first-decision time. Compaction
  prunes the snapshot with the message — it never migrates into m[0]/m[1]
  or a newer user message.
- **Persistence**: new independent `session_meta.subagent_inject_decisions`
  namespace (migration v85) with CAS / first-writer-wins, keyed by stable
  SessionEntry ID. Winning decisions persist the complete rendered payload;
  deterministic failures (invalid marker, capacity refusal, zero available)
  persist as explicit outcomes; transient faults persist nothing and retry —
  a retryable DB error never masquerades as "ID not found".
- **Capacity**: bypassing the memory budget does not bypass the model's hard
  context window — the payload is checked pre-submit against the live
  sane-bounded window and an oversized payload fails explicitly instead of
  truncating.
- **Idempotency**: owned by the persisted decision, never by text sniffing —
  a task that literally discusses `<ctx-subagent-inject>` still injects. The
  only append guard is an exact-payload match against same-pass double
  writes.
- **Ordering**: runs after the auto-search hint (first-time and replay wire
  order is always hint-then-snapshot); the new tag joins the stacked-
  augmentation detection list so an injected message never gets a fresh
  auto-search hint.

## Configuration

Single knob, default on, independent of `memory.enabled`:

```jsonc
{ "memory": { "subagent_inject": { "enabled": true } } }
```

## Scope (v1)

- **Pi only.** OpenCode is not implemented in this batch (the shared config
  schema carries the knob for both harnesses; only the Pi runtime acts on
  it).
- **Normal transform path only** — the compaction-off path is not supported
  in v1 (same gate as auto-search, documented).

## Measured results

- Hot path: **zero embedding calls, zero extra vector math** — plain
  `WHERE id IN (...)` by primary key; auto-search still runs normally when
  enabled.
- 34 new unit tests covering the matrix: parsing boundaries (including
  adversarial 5000-digit input bounded <200 ms), eligibility/no-metadata-
  leak, idempotent replay on reconstructed JSONL, snapshot freezing under
  later archive, CAS first-writer-wins, transient-fault retry without fake
  "not found", explicit capacity refusal, and string/multi-part/image-only
  message shapes.
- Full suites green: pi-plugin 1184 pass / 0 fail; plugin package 4800+
  pass (3 pre-existing parallel-mode perf flakes pass in isolation,
  unrelated).

## Why not alternatives

- **ctx_search**: semantic, fuzzy, embedding-bound — the opposite of "parent
  already knows the IDs".
- **Let the child call ctx_memory get**: requires tool-round-trip latency and
  the child's cooperation; the point is deterministic presence on the first
  request.
- **Visible-ID set filtering**: explicit naming exists precisely to recover
  budget-excluded memories, so already-visible IDs are not deduplicated
  away — the snapshot always renders.

G0-0000:
# Pi: task-requested memory injection via `⟦mc-mem: …⟧` markers (v1)

Closes #449

When a real user task message carries an explicit `⟦mc-mem: 11, 55⟧` marker,
fetch those memories by ID and append their full content to the end of that
message in a `<ctx-subagent-inject>` snapshot block — deterministic, sitewise,
zero-embedding, and outside the `injection_budget_tokens` budget. The agent
never needs to call a tool; the primary scenario is an orchestrator that
already knows exactly which memory IDs a subagent task needs (pi-subagents
children share the project memory store but arrive with no parent
conversation).

```text
Please follow the deploy SOP ⟦mc-mem: 11, 55⟧
```

```xml
<ctx-subagent-inject>
Task-requested memory snapshot (requested IDs: 11, 55; first read at 2026-09-14T10:00:00.000Z).
- [ID 11] [PROJECT_RULES] deploys happen only via the staging gate ...
- [ID 55] [CONFIG_VALUES] dashboard token budget is 8000 ...
Unavailable: 99 (not available).
</ctx-subagent-inject>
```

## Design decisions

| Decision | Rationale |
|---|---|
| Bounded single-line coarse locator + strict per-token validation | The naive regex has catastrophic backtracking (>1500 ms at 26 digits, measured). The fixed-character-class pattern is linear on any input; strict grammar (positive safe ints, optional `#`, space/tab/ASCII-comma separators) abandons malformed markers wholesale — never partially parsed. |
| Recognition limited to real user task text | Fenced/inline code, blockquotes, prior augmentation blocks, assistant/tool/synthetic messages are ignored; cross-part markers are never stitched; memory bodies in our own payload are never re-parsed. Literal discussion of the tag in a task must not trigger anything. |
| Message-level 10-ID cap (constant) | Guardrail against both query cost and token volume; over-limit IDs are reported as unprocessed, never silently dropped or back-filled. |
| Two-layer read filter | Access scope reuses the existing visibility rules via a narrowly extracted shared helper (`memory/visibility-filter.ts`, byte-identical to the former tool closure, which now delegates). Injection eligibility adds `active`/`permanent` + unexpired. Unauthorized IDs render one opaque note — no metadata leak. |
| Independent `subagent_inject_decisions` namespace (migration v85) | Never shares a decision slot with auto-search; CAS / first-writer-wins keyed by stable SessionEntry ID; winning decisions persist the complete rendered payload and replay it byte-for-byte; deterministic failures persist explicitly; transient faults persist nothing and retry. |
| Snapshot semantics | First read freezes availability + timestamp; later update/archive/expire does not refresh or retract history; compaction prunes with the message and never migrates into m[0]/m[1] or a newer message. |
| Pre-submit capacity check | Bypassing the memory budget must not overflow the model window: payload is checked against the live sane-bounded window; oversized fails explicitly — never silently truncated. |
| Exact-payload append guard only | Idempotency belongs to the persisted decision, never to substring-sniffing the bare tag (a task may discuss `<ctx-subagent-inject>` as plain text). Append-before-persist with rollback keeps half-persisted states impossible; contention realigns the wire to the winner's bytes. |
| Append to the LAST text part | The legacy helper appends to the first text part; §"message end" requires the last, preserving image/other parts. First pass and replay go through the same append path → byte-identical. |
| Mounted after auto-search | Wire order is always hint-then-snapshot, first-time and replay; the new tag joins `hasStackedAugmentation` so injected messages never get a fresh hint. |
| Single config knob, default on | `memory.subagent_inject.enabled` — synced across Zod schema, defaults, and the generated JSON schema; independent of `memory.enabled`. |

## Test evidence

- New `subagent-inject-pi.test.ts`: **34 cases** — parsing boundaries
  (single/multi markers, `#`, separators, dedupe, cap, LF/CRLF rejection,
  leading zeros/unsafe ints, glued tokens, full-width punctuation, region
  exclusion, cross-part non-stitching, 5000-digit adversarial input bounded
  <200 ms), eligibility & no-metadata-leak (archived/expired/foreign/missing),
  request-order restoration despite unordered SQL `IN`, closing-tag
  escaping, idempotent replay on reconstructed JSONL, snapshot freezing under
  later archive, non-tail no-decision discipline, unresolved-entry-id skip,
  transient DB fault retry without fake "not found", explicit capacity
  refusal, CAS first-writer-wins, and string/multi-part/image-only message
  shapes.
- Schema default/round-trip tests + migration fence tests updated for v85
  (incl. the armed-replay populate arm).
- `typecheck` / `lint` clean; pi-plugin suite **1184 pass / 0 fail**; plugin
  suite 4800+ pass (3 pre-existing parallel-mode perf flakes pass in
  isolation, unrelated to this change).

## v1 limitations (documented in CONFIGURATION.md)

- **Pi only** — OpenCode is not implemented in this batch.
- **Compaction-off path unsupported** — same gate as auto-search.
- Not gated on `isSubagent`: Pi cannot reliably see that flag, and the
  primary scenario (pi-subagents children running the full entry point) is
  exactly where markers appear in ordinary user task messages.

## Notes for reviewers

- `packages/plugin/src/tools/ctx-memory/tools.ts` delegates visibility to the
  new shared helper — behavior unchanged, logic byte-identical.
- The feature is deliberately **not** part of the memory budget or the
  visible-ID set: explicit naming exists to recover budget-excluded
  memories, and already-visible IDs still render in full beside the task.

---
<a href="https://app.blacksmith.sh/cortexkit/codesmith/magic-context/pr/450"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792010296&installation_model_id=22398&pr_number=450&repository=cortexkit%2Fmagic-context&return_to=https%3A%2F%2Fgithub.com%2Fcortexkit%2Fmagic-context%2Fpull%2F450&signature=85df89445d260a041e1d2d14bdd26b1010d88179933e97f7d4568462a60785d2"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

---
## Summary by cubic
Closes #449. Previously, Pi subagent tasks could not explicitly request known memories; now `⟦mc-mem: …⟧` markers fetch those memories by ID and append a full `<ctx-subagent-inject>` snapshot to the task.

**Behavior**

- Accepts up to 10 deduplicated positive IDs and abandons malformed markers instead of partially parsing them.
- Ignores markers in code, quoted, generated, assistant, tool, and prior augmentation content.
- Renders inaccessible, expired, archived, foreign, or missing IDs as one opaque unavailable note.
- Preserves message parts by appending to the last text part after auto-search output.
- Bypasses the memory injection budget but refuses payloads that exceed the live model window instead of truncating them.
- Uses Pi’s `tokenizePiMessages` for capacity checks, including thinking, signatures, tool calls, tool results, and fixed image accounting.
- Keeps auto-search retries working when an injection snapshot follows a transient search failure.

**Persistence and rollout**

- Freezes the first snapshot and replays it byte-for-byte through the `subagent_inject_decisions` namespace with CAS first-writer-wins behavior.
- Disabling `memory.subagent_inject.enabled` stops new injections but does not suppress persisted snapshots.
- Adds migration v85, removal cleanup, shared visibility filtering, and a default-on Pi-only configuration setting.
- OpenCode and the compaction-off path remain unsupported.
- Adds 42 injection tests covering parsing, capacity, wire shapes, eligibility, replay, contention, and lifecycle behavior.

<sup>Written for commit 1b8482dd214227f3c337b77302b2944e4777667b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/cortexkit/magic-context/pull/450?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Pi now reads explicit `⟦mc-mem: …⟧` markers in user tasks and appends the named memories as a frozen snapshot. The feature also persists replay decisions, checks model capacity, and adds a default-on configuration setting.

- Strictly parses markers and filters requested memories.
- Replays saved snapshots in stable message order and cleans them up with their messages.
- Wires the feature into Pi configuration, storage migrations, and documentation.
</details>

The latest capacity fix looks correct, but the open retry-order bug still blocks merging.

A retried search hint can still be appended after an existing memory snapshot because `hasStackedAugmentation` ignores that snapshot. The later replay sees the exact snapshot bytes and leaves them before the hint.

**Files Needing Attention:** `packages/pi-plugin/src/auto-search-pi.ts`

| Filename | Overview |
|----------|----------|
| packages/pi-plugin/src/subagent-inject-pi.ts | `estimateMessagesTokens` now uses Pi's shared message counter. This fixes the earlier drift for images and signatures. |
| packages/pi-plugin/src/subagent-inject-pi.test.ts | The new tests cover images, thinking, signatures, tool calls, and tool results. |
| packages/pi-plugin/src/auto-search-pi.ts | A retried search hint can still remain after a saved memory snapshot. The existing review thread remains open. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    actor User
    participant Pi
    participant Context as Context handler
    participant Search as Auto-search
    participant Inject as Memory injector
    participant DB as Context database
    User->>Pi: Send task with mc-mem marker
    Pi->>Context: Build outgoing messages
    Context->>Search: Add or replay search hint
    Search->>DB: Read or save hint decision
    Context->>Inject: Replay saved snapshots
    Inject->>DB: Read injection decisions
    alt New valid marker
        Inject->>DB: Read visible memories
        Inject->>Inject: Count current Pi messages
        alt Snapshot fits
            Inject->>DB: Save first decision
            Inject-->>Context: Append frozen snapshot
        else Snapshot does not fit
            Inject->>DB: Save capacity-exceeded
            Inject-->>Context: Leave task unchanged
        end
    end
    Context-->>Pi: Return final messages
    Pi-->>User: Run model turn
```
</details>

<sub>Reviews (4): Last reviewed commit: ["pi: count the capacity baseline with Pi&#39;..."](https://github.com/cortexkit/magic-context/commit/1b8482dd214227f3c337b77302b2944e4777667b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=64133705)</sub>


G0-0000:
Thanks for the review — every item was verified against the code. All findings are addressed in `4f623ae4` (pushed). Disposition per comment:

**P1 — capacity guard uses pre-transform usage (context-handler.ts)**
True. The guard now measures the outgoing wire: `estimateTokens` over all messages at our mount point (same-pass auto-search hints and note nudges included), checked as `wire + payload > sane-bounded window`. The handler passes only `contextLimit`; the stale `currentInputTokens` input is gone.

**P1 — concurrent no-inject winner leaves the local snapshot on the wire (subagent-inject-pi.ts)**
True. The reconciliation branch now strips the locally appended payload whenever the committed winner differs — including a `no-inject` winner, which previously fell through with our bytes stranded on the wire.

**P1 — markers inside generated hint/instruction blocks honored as user requests (subagent-inject-pi.ts scanner)** *(additional finding from the same review run, not in the summary list)*
True. `<ctx-search-hint>`, `<ctx-search-auto>`, and `<instruction>` blocks are now stripped before marker scanning, so a memory fragment quoted inside a generated block can never re-enter as a task request. Regression test added.

**P1 — `enabled: false` also kills replay, contradicting the documented contract (context-handler.ts gate)** *(additional finding from the same review run)*
True. The handler mount no longer gates on `enabled`; the runner always replays persisted snapshots and gates only NEW first-decisions. Disabling stops new injections without retracting history, exactly as the config docs describe.

**P2 — persistence throw after append strands bytes and can duplicate on retry (subagent-inject-pi.ts)**
True. Decision persistence is now inside a try/catch that rolls the wire bytes back and persists nothing, so a reused pass cannot append a second snapshot; CAS-exhausted gets the same rollback. Both retry next pass (transient semantics).

**P2 — auto-search retry permanently swallowed by the stacked check (auto-search-pi.ts)**
True. `<ctx-subagent-inject>` is removed from `hasStackedAugmentation`: the snapshot lands after auto-search ran (or after a retryable timeout), so counting it stacked recorded a permanent `no-hint: stacked` for messages whose search merely timed out once. `extractUserPromptText` now also strips the snapshot so an embedding retry sees the original prompt. New test simulates a timeout-then-recover pass pair.

**P2 — `removeSubagentInjectDecisionByMessageId` never wired into message removal (storage cleanup)**
True. Wired into `cleanupRemovedMessageState` alongside the note-nudge and auto-search removal helpers, with matching session logs.

**P2 — over-limit test assertion never fires (subagent-inject-pi.test.ts)**
True. Injected lines end with the bare content, so the trailing-`"]"` pattern was vacuous. Asserts on `body-10`/`body-11` bare now.

**P3 — CONFIGURATION.md ordering/suppression contradiction**
Fixed in both sections: the `subagent_inject` ordering paragraph and the `auto_search` suppression rules now state that a snapshot does NOT suppress the hint and explain why.

Tests: +6 new regression cases (38 in `subagent-inject-pi.test.ts`, 15 in `auto-search-pi.test.ts`). `typecheck`/`lint` clean; pi-plugin suite 1189 pass / 0 fail.

G0-0000:
Second review round addressed in `43ca7564` (pushed). All three findings were valid:

**P1 — capacity baseline omits images/thinking/tool calls/tool results**
Fixed. `estimateMessagesTokens` now walks every Pi wire shape: assistant thinking parts, toolCall name+arguments, tool-result text, and image base64 payloads (the old top-level `reasoning` probe was dead code for this SDK — thinking lives in content parts). Image base64 is counted as transmitted: it over-estimates versus provider per-image pricing, which is the safe direction for a guard that must refuse rather than allow an overflow. New regression test: a marker task beside a high-entropy base64 image is refused at a window the text-only count would have allowed, and injects at a generous one.

**P2 — user-authored `<instruction>` blocks stripped too broadly**
Fixed with the suggested regex verbatim: the exclusion now matches only `<instruction name="deferred_notes">` (the generated note-nudge block). Markers inside user-authored instruction blocks are honored as task text — consistent with the contract that only generated/historical/augmentation blocks are ignored. New test covers both directions.

**P2 — CAS-contention test never reached the contention path**
Correct, and a good catch: the pre-persisted winner made the runner early-return before append/CAS, so the strip-on-no-inject-winner branch was untested. Adopted the suggested test — winner hidden via `decisions: []`, `appendSubagentInjectDecision` mocked to return the already-present no-inject decision — and kept the old variant separately as the already-decided early-return case.

Tests: 41 in `subagent-inject-pi.test.ts`; `typecheck`/`lint` clean; pi-plugin suite 1192 pass / 0 fail.

G0-0000:
Third review round addressed in `1b8482dd` (pushed). Both findings were valid and are resolved by removing the divergence entirely:

**P1 — capacity count inconsistent with Pi's own accounting (both directions)**
Fixed by delegation, not by tuning our own walk: the pre-submit capacity baseline now comes from `tokenizePiMessages` — the same counter `/ctx-status` and the compaction budget use. That settles both halves of the finding structurally: signatures (`textSignature`/`thinkingSignature`) are counted because Pi counts them, and images use Pi's fixed 1200-token OpenCode-compatible fallback, so the base64 over-estimation false-refusals are gone too. "What we count" and "what Pi counts" can no longer drift apart. The payload itself is still estimated with `estimateTokens` (a small, additive term).

**P3 — test name claimed thinking/tool-call coverage the message never exercised**
Correct. The media test is renamed to what it actually covers (fixed image accounting, which now needs no payload trickery). A new test builds an assistant message with a thinking part + signature and a toolCall with arguments, plus a toolResult body, so those tokenizer branches genuinely execute: the wire refuses at a window the marker text alone would pass, and the decision records `capacity-exceeded`.

Tests: 42 in `subagent-inject-pi.test.ts`; `typecheck`/`lint` clean; pi-plugin suite 1193 pass / 0 fail.

G0-0000:
Thanks for the detailed feedback and for running the tests. I understand and accept the decision.

To clarify my intent: I did not open this PR expecting it to be merged as-is. I used it mainly as a carrier for the prototype code, tests, and diff, so the idea could be concrete and verifiable. The Pi-only scope, the default-on knob, the marker syntax, and the schema migration were prototype choices, not a request to approve that shape. I also understand now that opening #449 alongside the PR is not the design approval step.

I will stop pushing this PR. I'll continue the design discussion in #449 and follow the maintainers' direction on the cross-harness contract. I don't have the full picture across OpenCode, Pi/OMP, and the Rust lane, so I won't try to define that framework myself. If it's useful, I can keep the prototype as reference material or adapt it to whatever contract you define.

Should I close this PR or convert it to draft?

G0-0000:
One clarification: I'm not dropping the use case, just the current PR shape. I'll continue the design discussion in #449 and am willing to adapt the prototype to whatever contract you define.

